PR #42, opened 2019-07-15
Improve logging and other minor fixes

### LINKED ISSUE #41: Layer log messages are too vague
Fossilize layer log messages are too vague. I got this particular error message `Fossilize ERROR: Handle is not registered.` multiple times today. It would be helpful if it specified which handle is not registered (e.g. pipeline handle, render pass handle...). Even the word registered is too vague - the error message could give a hint why the handle "registration" fails.

HansKristian-Work:
- Also, use a more portable PRIx64 format identifier (MinGW and C99 printf formats like %llx is dicey).
- Fixup fossilize-opt to use Fossilize DB format.
- Update submodules.

Fix #41.

### DIFF
--- a/cli/fossilize_bench.cpp
+++ b/cli/fossilize_bench.cpp
@@ -27,6 +27,7 @@
 #include <memory>
 #include <random>
 #include <vector>
+#include <inttypes.h>
 
 using namespace Fossilize;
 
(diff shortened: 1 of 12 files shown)


jammymalina:
@HansKristian-Work Nice job!

PR #44, opened 2019-08-26
Add support to VK_KHR_pipeline_executable_properties

ibriano:
I made the choice of going with a semi-handcrafted JSON file for the intermediary stats files, as it made it a lot easier to identify them later and parse them into one thing from which to dump all the stats to a CSV.

### DIFF
--- a/cli/device.cpp
+++ b/cli/device.cpp
@@ -244,7 +244,15 @@ bool VulkanDevice::init_device(const Options &opts)
 	VkPhysicalDevicePipelineExecutablePropertiesFeaturesKHR stats_feature = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PIPELINE_EXECUTABLE_PROPERTIES_FEATURES_KHR };
 	gpu_features2.pNext = &stats_feature;
 	if (has_device_features2)
+	{
 		vkGetPhysicalDeviceFeatures2(gpu, &gpu_features2);
+
+		pipeline_stats = stats_feature.pipelineExecutableInfo;
+		if (pipeline_stats && opts.want_pipeline_stats)
+			stats_feature.pipelineExecutableInfo = VK_TRUE;
+		else
+			stats_feature.pipelineExecutableInfo = VK_FALSE;
+	}
 	else
 		vkGetPhysicalDeviceFeatures(gpu, gpu_features);
 
--- a/cli/device.cpp
+++ b/cli/device.cpp
@@ -236,17 +236,24 @@ bool VulkanDevice::init_device(const Options &opts)
 	}
 
 	api_version = target_api_version;
+	bool has_device_features2 = api_version >= VK_MAKE_VERSION(1, 1, 0);
 
 	// FIXME: There are arbitrary features we can request here from physical_device_features2.
-	VkPhysicalDeviceFeatures gpu_features = {};
-	vkGetPhysicalDeviceFeatures(gpu, &gpu_features);
+	VkPhysicalDeviceFeatures2 gpu_features2 = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2 };
+	VkPhysicalDeviceFeatures *gpu_features = &gpu_features2.features;
+	VkPhysicalDevicePipelineExecutablePropertiesFeaturesKHR stats_feature = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PIPELINE_EXECUTABLE_PROPERTIES_FEATURES_KHR };
+	gpu_features2.pNext = &stats_feature;
+	if (has_device_features2)
+		vkGetPhysicalDeviceFeatures2(gpu, &gpu_features2);
+	else
+		vkGetPhysicalDeviceFeatures(gpu, gpu_features);
 
 	// FIXME: Have some way to enable the right features that a repro-capture may want to use.
 	// FIXME: It is unlikely any feature other than robust access has any real impact on code-gen, but who knows.
-	if (gpu_features.robustBufferAccess && opts.features && opts.features->features.robustBufferAccess)
-		gpu_features.robustBufferAccess = VK_TRUE;
+	if (gpu_features->robustBufferAccess && opts.features && opts.features->features.robustBufferAccess)
+		gpu_features->robustBufferAccess = VK_TRUE;
 	else
-		gpu_features.robustBufferAccess = VK_FALSE;
+		gpu_features->robustBufferAccess = VK_FALSE;
 
 	// Just pick one graphics queue.
 	// FIXME: Does shader compilation depend on which queues we have enabled?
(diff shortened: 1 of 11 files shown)


cli/device.cpp
```diff
@@ -238,15 +238,17 @@ bool VulkanDevice::init_device(const Options &opts)
 	api_version = target_api_version;
 
 	// FIXME: There are arbitrary features we can request here from physical_device_features2.
-	VkPhysicalDeviceFeatures gpu_features = {};
-	vkGetPhysicalDeviceFeatures(gpu, &gpu_features);
+	VkPhysicalDeviceFeatures2 gpu_features = {};
```
HansKristian-Work: This will only work on Vulkan 1.1 devices, please add fallbacks for Vulkan 1.0.

fossilize_external_replayer.hpp
```diff
@@ -51,6 +51,13 @@ class ExternalReplayer
 		// Maps to --device-index.
 		unsigned device_index;
 
+		// Carve out a range of which pipelines to replay.
+		// Used for multi-process replays where each process gets its own slice to churn through.
```
HansKristian-Work: For cross-ABI concerns, this struct needs to be a POD, without an implicit constructor.. Consider adding a bool to flag if these fields should be used.

cli/fossilize_replay_windows.hpp
```diff
@@ -640,14 +640,31 @@ static int run_master_process(const VulkanDevice::Options &opts,
 	vector<ProcessProgress> child_processes(processes);
 	vector<HANDLE> wait_handles;
 
+	size_t requested_graphic_pipelines = replayer_opts.end_graphics_index - replayer_opts.start_graphics_index;
```
HansKristian-Work: Is this range checked? Seems like it can easily go to UINT32_MAX without being clamped against real size of archive.

fossilize_external_replayer_linux.hpp
```diff
@@ -368,6 +368,20 @@ bool ExternalReplayer::Impl::start(const ExternalReplayer::Options &options)
 		sprintf(index_name, "%u", options.device_index);
 		argv.push_back(index_name);
 
+		if (options.start_graphics_index != 0 || options.end_graphics_index != ~0u)
+		{
+			argv.push_back("--graphics-pipeline-range");
+			argv.push_back(std::to_string(options.start_graphics_index).c_str());
```
HansKristian-Work: std::string goes out of scope here and UB.

fossilize_db.cpp
```diff
@@ -233,6 +233,14 @@ struct DumbDirectoryDatabase : DatabaseInterface
 		return true;
 	}
 
+	const char * get_db_path_for_hash(ResourceTag tag, Hash hash) override
```
HansKristian-Work: `const char *get_db_path_for_hash`

cli/device.cpp
```diff
@@ -250,6 +252,12 @@ bool VulkanDevice::init_device(const Options &opts)
 	else
 		gpu_features.features.robustBufferAccess = VK_FALSE;
 
+	pipeline_stats = stats_feature.pipelineExecutableInfo;
```
HansKristian-Work: Extension is not checked for. The extension should not be enabled when want_pipeline_stats is false, similar to VK_AMD_shader_info. At least that extension would force-disable on-disk pipeline cache if the extension was just enabled, which is a serious performance concern when replaying.

cli/fossilize_replay.cpp
```diff
@@ -419,6 +427,115 @@ struct ThreadedReplayer : StateCreatorInterface
 		return true;
 	}
 
+	void get_pipeline_stats(ResourceTag tag, Hash hash, VkPipeline pipeline)
+	{
+		VkPipelineInfoKHR pipeline_info;
```
HansKristian-Work: `pipeline_info = { VK_STRUCTURE_TYPE ... }` is used throughout the codebase to init structs.

cli/fossilize_replay.cpp
```diff
@@ -1952,6 +2099,157 @@ static int run_progress_process(const VulkanDevice::Options &device_opts,
...
+		return false;
+	}
+
+	fread(buffer, fsize, 1, json_fp);
+	fclose(json_fp);
+
+	doc.Parse(buffer, fsize);
+	bool ret = !doc.HasParseError();
+
+	delete buffer;
```
HansKristian-Work: `delete[]`

cli/fossilize_replay.cpp
```diff
@@ -1952,6 +2099,157 @@ static int run_progress_process(const VulkanDevice::Options &device_opts,
...
+	fseek(json_fp, 0, SEEK_END);
+	long fsize = ftell(json_fp);
+	if (fsize <= 0)
+	{
+		fclose(json_fp);
+		return false;
+	}
+	fseek(json_fp, 0, SEEK_SET);
+
+	char *buffer = new char[fsize];
```
HansKristian-Work: Just use std::vector

cli/fossilize_replay.cpp
```diff
@@ -1952,6 +2099,157 @@ static int run_progress_process(const VulkanDevice::Options &device_opts,
...
+	{
+		colnumber = 0;
+		for (size_t i = 0; i < header.size(); i++)
+		{
+			auto col = r.find(i);
+			fprintf(fp, "%s%s", col == r.end() ? "" : col->second.c_str(), ++colnumber < header.size() ? "," : "\n");
+		}
+	}
+
+	fclose(fp);
```
HansKristian-Work: Are the temporary JSON files cleaned up at the end?

cli/fossilize_replay.cpp
```diff
@@ -1952,6 +2099,157 @@ static int run_progress_process(const VulkanDevice::Options &device_opts,
...
+
+	delete buffer;
+
+	return ret;
+}
+
+static void stats_to_csv(std::string stats_path, rapidjson::Document &doc)
+{
+	std::vector<std::string> header;
+	std::map<std::string, size_t> columns;
```
HansKristian-Work: Would std::unordered_map work?

cli/fossilize_replay.cpp
```diff
@@ -419,6 +427,115 @@ struct ThreadedReplayer : StateCreatorInterface
...
+				execs.PushBack(pe, alloc);
+			}
+			doc.AddMember("executables", execs, alloc);
+
+			rapidjson::StringBuffer buffer;
+			rapidjson::Writer<rapidjson::StringBuffer> writer(buffer);
+			doc.Accept(writer);
+
+			{
+				lock_guard<mutex> lock(pipeline_stats_queue_mutex);
```
HansKristian-Work: Why take a lock here?

HansKristian-Work: Oh, right. Each thread opens, writes and closes the same file.

HansKristian-Work:
Merged, thanks! I made some changes on top of your branch, biggest change was replaying the appends of the JSON file to using a Fossilize archive, which should make Windows build not crawl. Some other minor nits as well, but not worth failing review for.

PR #65, opened 2020-03-20
Miscellaneous changes

pendingchaos:
Not too certain about the device recreation change. It can fix issues where the replay tool is given multiple databases and ends up requesting too old of a Vulkan version for one of them. I split the commit into two to make review easier.

### DIFF
--- a/cli/fossilize_feature_filter.cpp
+++ b/cli/fossilize_feature_filter.cpp
@@ -285,19 +285,29 @@ bool FeatureFilter::Impl::pnext_chain_is_supported(const void *pNext) const
 		switch (base->sType)
 		{
 		case VK_STRUCTURE_TYPE_PIPELINE_TESSELLATION_DOMAIN_ORIGIN_STATE_CREATE_INFO:
-			return enabled_extensions.count(VK_KHR_MAINTENANCE2_EXTENSION_NAME) != 0 || api_version >= VK_API_VERSION_1_1;
+			if (!enabled_extensions.count(VK_KHR_MAINTENANCE2_EXTENSION_NAME) && api_version < VK_API_VERSION_1_1)
+				return false;
+			break;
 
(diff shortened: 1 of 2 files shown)


pendingchaos:
I've created a separate PR for the validation checker fixes: https://github.com/ValveSoftware/Fossilize/pull/66

When running the replay tool with several databases (Sascha Willems demos and pipelines from vkpipeline-db's runner), it seems it would end up using the vulkan version of one the Sascha Willems demos (Vulkan 1.0) but some of the vkpipeline-db pipelines require Vulkan 1.1

It's not a real issue at the moment, since statistics from Sascha Willems demos aren't too interesting. If it every becomes a real issue, I think it should be possible to run the replay tool once for each application instead of all at once

HansKristian-Work:
Is this happening when using the "combined" multi-archive support of fossilize-replay?

pendingchaos:
> Is this happening when using the "combined" multi-archive support of fossilize-replay?

I pass several databases to fossilize-replay (like `./fossilize-replay --enable-pipeline-stats output.csv sascha-willems/bloom.foz rottr.foz ...`). If that's what it's called, then yes

HansKristian-Work:
Right, that approach is somewhat fiddly w.r.t. application infos. The recommended approach is to replay different foz archives separately in different processes for this case, but one thing we can add to Fossilize is to look at all application infos first, and then select which one we should be using based on things like Vulkan version or other user-defined filters.

pendingchaos:
> Right, that approach is somewhat fiddly w.r.t. application infos. The recommended approach is to replay different foz archives separately in different processes for this case, but one thing we can add to Fossilize is to look at all application infos first, and then select which one we should be using based on things like Vulkan version or other user-defined filters.

It shouldn't be too difficult to update the replay script I'm using to run fossilize-replay for each application and concatenate the CSVs. Since that's the recommended approach, I think I'll do that

Should I close this PR?

HansKristian-Work:
Sorry for late reply. Yes, I think this can be closed. The appinfo selection should be solved, but in a very different way. I'll look into that eventually.

PR #66, opened 2020-03-20
Validation fixes

pendingchaos:


cli/fossilize_feature_filter.cpp
```diff
@@ -393,8 +407,10 @@ bool FeatureFilter::Impl::pnext_chain_is_supported(const void *pNext) const
 		case VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_REQUIRED_SUBGROUP_SIZE_CREATE_INFO_EXT:
 		{
 			// Should correlate with stage.
-			return features.subgroup_size_control.subgroupSizeControl == VK_FALSE &&
-			       props.subgroup_size_control.requiredSubgroupSizeStages != 0;
+			if (features.subgroup_size_control.subgroupSizeControl == VK_TRUE ||
```
HansKristian-Work: Actually, hm, shouldn't this be `if (test == VK_FALSE ||)`? Was probably a bug in first implementation.

pendingchaos: Yeah, I think so

I'll change it

HansKristian-Work:
Tested build locally, so I'll just merge now.

PR #67, opened 2020-03-28
fossilize_replay: Add a --pipeline-hash option

jekstrand:
This option allows the user to deterministically select exactly one
pipeline from the database.  This is very useful when using fossilize to
examine compiler behavior of individual shaders.

### DIFF
--- a/cli/fossilize_replay.cpp
+++ b/cli/fossilize_replay.cpp
@@ -231,6 +231,9 @@ struct ThreadedReplayer : StateCreatorInterface
 
 		unsigned shader_cache_size_mb = 256;
 
+		// Hash for replaying a single pipeline
+		Hash pipeline_hash = 0;
+
 		// Carve out a range of which pipelines to replay.
 		// Used for multi-process replays where each process gets its own slice to churn through.
 		unsigned start_graphics_index = 0;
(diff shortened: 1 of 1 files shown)


cli/fossilize_replay.cpp
```diff
@@ -231,6 +231,8 @@ struct ThreadedReplayer : StateCreatorInterface
 
 		unsigned shader_cache_size_mb = 256;
 
+		const char *pipeline_hash = nullptr;
```
HansKristian-Work: This should be std::string, and not part of the ThreadedReplayer::Options since it does not affect the ThreadedReplayer.

HansKristian-Work: Alternatively, the Hash uint64_t can be parsed directly when parsing arguments, no need to pass around a string.

HansKristian-Work: Looking at it some more, just keep the pipeline_hash option here, and change the type to Hash directly.

jekstrand: > Alternatively, the Hash uint64_t can be parsed directly when parsing arguments, no need to pass around a string.

I thought about that but I couldn't figure out how to properly propagate the error from the parser lambda.  I guess I could call exit()?

HansKristian-Work: Log and exit() is fine with me.

cli/fossilize_replay.cpp
```diff
@@ -2994,6 +3027,16 @@ int main(int argc, char *argv[])
 	});
 	cbs.add("--num-threads", [&](CLIParser &parser) { replayer_opts.num_threads = parser.next_uint(); });
 	cbs.add("--loop", [&](CLIParser &parser) { replayer_opts.loop_count = parser.next_uint(); });
+	cbs.add("--pipeline-hash", [&](CLIParser &parser) {
+		const char *hash_str = parser.next_string();
+		char *end;
+		replayer_opts.pipeline_hash = strtoll(hash_str, &end, 16);
```
HansKristian-Work: Actually, shouldn't this should be strtoull?

jekstrand: Yes, it should.  Good catch!  Fixed in the latest version.

PR #68, opened 2020-04-11
⚡ Move most conditional logic to generator expressions (and other patches)

slurps-mad-rips:
This PR reduces configure time to a decent degree. While generator expressions can be painful to look at in some cases, I tried to keep them simple, minus one area they were needed due to the older version of CMake this project targets. If you're curious to know more about CMake generator expressions, I recommend looking at the documentation for them. That said they are a bit undiscussed but I can easily summarize them as "What if Lisp was bad, had awful syntax, and didn't support spaces in between arguments?".

That's CMake's generator expressions. 

They can also execute in parallel at project generation time so it reduces the amount of time CMake is running at configure time, which is a single threaded operation that executes on an AST, versus the semi-compiled multi-threaded approach project generation can take.

This patch also:

1) Attempts to fix some warnings for newer versions of CMake
2) Uses the `CMAKE_<LANG>_VISIBILITY_PRESET` builtin variable (hence `CMP0063` being set to `NEW`.
3) Permits compiling fossilize under newer versions of C++ and C if used as a subproject
4) Uses the Threads package to use pthreads as possible. This means that Android will still work, but compiler details are not needed
5) Moved most options to the top of the file for folks who don't use CMake GUI or ccmake so they can more easily judge some of the options provided
6) Disable running tests UNLESS `BUILD_TESTING` is on *AND* fossilize is the root project
7) Users can not provide their own miniz library if they would like.
8) Left some notes for further improvements regarding MSVC and sanitizers
9) Fixed an issue with MinGW and import libraries causing filesystem race conditions. This is a temporary fix and more work is needed
10) Use `target_link_options` in the root CMakeLists file when possible. Fall back to setting properties otherwise. This does admittedly need improvement.

A future PR will contain a fossilize_link_options command that properly
sets LINK_OPTIONS when possible, and LINK_FLAGS on pre CMake 3.13
versions.

left a note to eventually tackle the `/Ehsc-` problem with MSVC. I can
handle this later, but it requires some additional work and dependent
projects need a massive improvement as well. Fixing fossilize will not
fix them, and I'd like to tackle them as well. Just because we're using
CMake doesn't mean we *all* have to suffer. (Instead, you can just let
me suffer 🤣)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,54 +1,102 @@
 cmake_minimum_required(VERSION 3.5)
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_C_STANDARD 99)
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
 project(Fossilize LANGUAGES CXX C)
 
-option(FOSSILIZE_MSVC_ANALYZE "Enable /analyze" OFF)
-option(FOSSILIZE_ENABLE_EXCEPTIONS "Compile Fossilize with exception support." OFF)
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+set(CMAKE_THREA_PREFER_PTHREAD ON)
 
-if (CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang"))
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fvisibility=hidden")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-    if (NOT FOSSILIZE_ENABLE_EXCEPTIONS)
-        set(FOSSILIZE_EXCEPTION_FLAGS "-fno-exceptions")
-    else()
-        set(FOSSILIZE_EXCEPTION_FLAGS "")
-    endif()
-    set(FOSSILIZE_CXX_FLAGS -Wall -Wextra -Wunused-result -Wno-missing-field-initializers -Wno-empty-body -Wshadow ${FOSSILIZE_EXCEPTION_FLAGS})
-elseif (MSVC)
-    if (FOSSILIZE_MSVC_ANALYZE)
-        set(FOSSILIZE_ANALYZE_FLAGS "/analyze")
-    else()
-        set(FOSSILIZE_ANALYZE_FLAGS "")
-    endif()
-    # Disabling exceptions in MSVC without a ton of scary warnings is quite finicky.
-    # Don't bother. We can verify no exceptions are present with -fno-exceptions on GCC/Clang.
-    set(FOSSILIZE_CXX_FLAGS ${FOSSILIZE_ANALYZE_FLAGS} /D_CRT_SECURE_NO_WARNINGS /D_SCL_SECURE_NO_WARNINGS /wd4267 /wd4244 /wd4309 /wd4005 /MP /DNOMINMAX)
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+include(CTest)
+
+find_package(Threads REQUIRED)
+find_package(Vulkan)
+
+if (NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
+  set(CMAKE_CXX_STANDARD 14)
 endif()
+if (NOT DEFINED CMAKE_C_STANDARD OR CMAKE_C_STANDARD LESS 99)
+  set(CMAKE_C_STANDARD 99)
+endif()
+
+if (NOT DEFINED CMAKE_CXX_VISIBILITY_PRESET)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+endif()
+
+if (NOT DEFINED CMAKE_C_VISIBILITY_PRESET)
+  set(CMAKE_C_VISIBILITY_PRESET hidden)
+endif()
+
+cmake_dependent_option(FOSSILIZE_TESTS
+  "Enable Fossilize tests" ON
+  "CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME;BUILD_TESTING" OFF)
+
+option(FOSSILIZE_ENABLE_EXCEPTIONS "Compile Fossilize with exception support." OFF)
+option(FOSSILIZE_MSVC_ANALYZE "Enable /analyze" OFF)
+# Not needed for position-independent-code, as its always turned on
+option(FOSSILIZE_VULKAN_LAYER "Build Vulkan layer." ON) 
+option(FOSSILIZE_CLI "Enable Fossilize CLI support." ON)
 
 option(FOSSILIZE_SANITIZE_ADDRESS "Sanitize address" OFF)
-set(FOSSILIZE_LINK_FLAGS)
-if (FOSSILIZE_SANITIZE_ADDRESS)
-    set(FOSSILIZE_CXX_FLAGS ${FOSSILIZE_CXX_FLAGS} -fsanitize=address)
-    set(FOSSILIZE_LINK_FLAGS "${FOSSILIZE_LINK_FLAGS} -fsanitize=address")
-endif()
-
 option(FOSSILIZE_SANITIZE_THREADS "Sanitize threads" OFF)
-if (FOSSILIZE_SANITIZE_THREADS)
-    set(FOSSILIZE_CXX_FLAGS ${FOSSILIZE_CXX_FLAGS} -fsanitize=thread)
-    set(FOSSILIZE_LINK_FLAGS "${FOSSILIZE_LINK_FLAGS} -fsanitize=thread")
+
+# option() cannot technically be used in newer CMake versions for non-boolean values
+# warnings will be thrown and this can be annoying otherwise.
+set(FOSSILIZE_VULKAN_INCLUDE_PATH "${PROJECT_SOURCE_DIR}/khronos/vulkan"
+  CACHE PATH "include path for vulkan headers")
+set(FOSSILIZE_RAPIDJSON_INCLUDE_PATH "${PROJECT_SOURCE_DIR}/rapidjson/include"
+  CACHE PATH "include path for rapidjson headers")
+
+# TODO: Gate all of this behind a better generator expression
+if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    list(APPEND FOSSILIZE_CXX_FLAGS
+      -Wall -Wextra -Wunused-result -Wno-missing-field-initializers -Wno-empty-body -Wshadow
+      $<$<NOT:$<BOOL:${FOSSILIZE_ENABLE_EXCEPTIONS}>>:-fno-exceptions>)
+elseif (MSVC)
+    # XXX: This can be done, and safely, but will require more work and a more recent version of CMake
+    # Disabling exceptions in MSVC without a ton of scary warnings is quite finicky.
+    # Don't bother. We can verify no exceptions are present with -fno-exceptions on GCC/Clang.
+    list(APPEND FOSSILIZE_CXX_FLAGS
+      $<$<BOOL:${FOSSILIZE_MSVC_ANALYZE}>:/analyze>
+      /wd4267
+      /wd4244
+      /wd4309
+      /wd4005
+      /MP)
 endif()
 
-include(GNUInstallDirs)
+# TODO: This does not currently support the MSVC support for AddressSanitizer
+if (FOSSILIZE_SANITIZE_ADDRESS)
+    list(APPEND FOSSILIZE_CXX_FLAGS -fsanitize=address)
+    list(APPEND FOSSILIZE_LINK_FLAGS -fsanitize=address)
+endif()
 
-add_library(miniz STATIC miniz/miniz.h miniz/miniz.c)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,54 +1,102 @@
...
-option(FOSSILIZE_MSVC_ANALYZE "Enable /analyze" OFF)
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+set(CMAKE_THREA_PREFER_PTHREAD ON)
+
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+include(CTest)
+
+find_package(Threads REQUIRED)
+find_package(Vulkan)
```
HansKristian-Work, line 14: External Vulkan is not needed.

CMakeLists.txt
```diff
@@ -1,54 +1,102 @@
 cmake_minimum_required(VERSION 3.5)
-set(CMAKE_CXX_STANDARD 14)
-set(CMAKE_C_STANDARD 99)
+set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+set(CMAKE_POLICY_DEFAULT_CMP0063 NEW)
 project(Fossilize LANGUAGES CXX C)
 
-option(FOSSILIZE_MSVC_ANALYZE "Enable /analyze" OFF)
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+set(CMAKE_THREA_PREFER_PTHREAD ON)
```
HansKristian-Work, line 7: Typo?

CMakeLists.txt
```diff
@@ -1,54 +1,102 @@
...
+
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+include(CTest)
+
+find_package(Threads REQUIRED)
+find_package(Vulkan)
+
+if (NOT DEFINED CMAKE_CXX_STANDARD OR CMAKE_CXX_STANDARD LESS 14)
```
HansKristian-Work, line 16: Doesn't this have to be defined before project() to take affect?

CMakeLists.txt
```diff
@@ -1,54 +1,102 @@
...
-    if (NOT APPLE AND NOT ANDROID)
-        target_compile_definitions(miniz PRIVATE _LARGEFILE64_SOURCE)
-    endif()
+if (NOT TARGET miniz)
+  add_library(miniz STATIC miniz/miniz.h miniz/miniz.c)
+  set_target_properties(miniz PROPERTIES POSITION_INDEPENDENT_CODE ON)
+  target_include_directories(miniz
+    PUBLIC
+      $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/miniz>)
+  string(CONCAT largefile $<AND:
```
HansKristian-Work, line 86: This is unreadable.

CMakeLists.txt
```diff
@@ -63,75 +111,51 @@ add_library(fossilize STATIC
...
 else()
-    target_include_directories(fossilize PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/rapidjson/include)
-endif()
-
-option(FOSSILIZE_VULKAN_INCLUDE_PATH "Optional custom include path for Vulkan headers.")
-if (FOSSILIZE_VULKAN_INCLUDE_PATH)
-    target_include_directories(fossilize PUBLIC ${FOSSILIZE_VULKAN_INCLUDE_PATH})
-else()
-    target_include_directories(fossilize PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/khronos/vulkan)
+  target_link_options(fossilize PRIVATE ${FOSSILIZE_LINK_FLAGS})
```
HansKristian-Work, line 148: This fails Travis.

CMakeLists.txt
slurps-mad-rips: I forgot to remove this, thanks for the catch. Fossilize doesn't work with stock Vulkan anyhow.

CMakeLists.txt
slurps-mad-rips: Yes, typo 😅

CMakeLists.txt
slurps-mad-rips: Surprisingly no! You can test this yourself manually via `get_property(standard TARGET <target-name> PROPERTY CXX_STANDARD)`. CMake uses `CMAKE_CXX_STANDARD` to initialize the `CXX_STANDARD` property on targets, but if it's left blank, it leaves it blank as well and lets the compiler's default kick in. This unfortunately causes problems if you just want to use `/std:latest` on MSVC (as some CMake versions don't understand it), but in practice it's rarely needed :)

CMakeLists.txt
slurps-mad-rips: I can break up some of the logic into something like `set(is-gnu-or-clang ...)`, but I personally find it no more readable than lisp. It can be easier to read if a newer CMake minimum version was used, but I'm unsure what that policy is.

slurps-mad-rips:
> How much do these optimizations actually matter?

As I said in my PR, it reduces configuration time. Using fossilize in a large build (as well as its dependencies) currently tanks configure time. Moving most logic to generation time improves throughput, as generation happens in a multi-threaded set of operations, versus the single thread only performance of configure. If you look at the internals for CMake, it creates a single generator expression evaluator and only uses that for operations, however within stuff like `if()`, it creates a whole interpreter *every* time you enter a branch, so it's best to reduce your `if()` expressions to guards or fallbacks for shims.

A majority of these changes are sadly intersectional. Breaking them up to improve fossilize as a library will be difficult, as a majority of the improvements are things like `$<BUILD_INTERFACE>`, relying on documented generator expressions versus undocumented variables like `CMAKE_CXX_COMPILER_ID` (which is technically not guaranteed to have the same 'name' as `$<CXX_COMPILER_ID>`, and has caused issues when moving between Clang on Windows (MSVC variant vs MinGW variant), AppleClang (the clang bundled with XCode), and Clang on macOS (built from source)

CMakeLists.txt
slurps-mad-rips: Is it this line in particular, or some other issue? I think I might have messed up the stringification here, which I can put into another commit to fix the above changes.

HansKristian-Work:
```
CMake Error at CMakeLists.txt:148 (target_link_options):

  Unknown CMake command "target_link_options".
```

CMakeLists.txt
HansKristian-Work: Minspec is 3.5 as the CMakeLists.txt declares. If it's possible to use some temporary variables here, that would probably make it more readable.

CMakeLists.txt
HansKristian-Work: Ok. As long as C++14 is enabled, I don't really mind how this is solved.

CMakeLists.txt
HansKristian-Work: By stock Vulkan, do you mean trying to link against Vulkan with the package?

HansKristian-Work:
So using `if()` in any way, shape or form tanks CMake? That is ... well. Surprising that Fossilize is the culprit since that sounds like a common pattern in all CMakeLists.txt I've ever seen. Anyways, if it truly does improve configure time, I can tolerate harder to read CMake.

CMakeLists.txt
slurps-mad-rips: The one I installed from LunarXChange, yes. Headers are not included as `<vulkan/vulkan.h>`, but rather as `"vulkan.h"`. There are also several headers missing from 'stock' Vulkan.

When you "link" against a library in CMake, it'll perform something known as transitive linking, where additional properties, such as `[INTERFACE_]INCLUDE_DIRECTORIES`, `COMPILE_OPTIONS`, etc. are used. The only time this doesn't occur is when the `TYPE`  of a target is `OBJECT`, where you need to do `$<TARGET_OBJECTS:${my-target}>` but this was fixed in a more recent version of CMake, though it requires setting a policy, and if your minimum is older it obviously won't have the intended effect. Even if I were to just do
`target_include_directories(fossilize PRIVATE $<TARGET_PROPERTY:Vulkan::Vulkan,INCLUDE_DIRECTORIES>)`, fossilize will fail to compile.

CMakeLists.txt
slurps-mad-rips: Totally understand. 🙂

CMakeLists.txt
slurps-mad-rips: Will do in my next commit. 🙂

CMakeLists.txt
slurps-mad-rips: I'll have to do some work to get this to work correctly. It shouldn't have entered the `else()` block and parsed this. Regardless, I needed to make a proper shim anyhow, and this was just a band-aid.

slurps-mad-rips:
Fossilize *by itself* is not the culprit, but think of it as "every CMakeLists.txt combined is the culprit". I found it easier to edit fossilize (because it's technically at the bottom of the dependency list) before I start tackling all of the SPIR-V/rapidjson CMakeLists.txt files (I also have to get around to LLVM's CMakeLists one of these days, but that's going to also require a proposal, approval from the core team, helping every author currently targeting LLVM, not breaking existing builds and honestly there's just not enough hours in the day 😭).

 Having a single project use a ton of `if()` calls isn't bad. Having a monolithic build that uses a lot of `if()` calls because you have 20 or so dependencies? That starts to slow stuff down as it compounds performance. Kind of like if every time CMake say an `if()` it allocated an entire lua VM just to do boolean logic, which is what it basically does (minus any actual Lua, of course 🤣). 

For reference, I've modified some CMake projects to use generator expressions over `if()` calls and the build time bottleneck started to become the compiler itself, rather than the configure portion.

(If you've ever compiled all of LLVM and Clang from source, you know how long the configure time can take, when it could be heavily reduced and most of your time waiting would be relegated to waiting for the actual build to complete 😬)

That said, I'll get to tackling this later this week, or possibly this weekend. Hobbies and free time and all that 😅

HansKristian-Work:
> Fossilize _by itself_ is not the culprit, but think of it as "every CMakeLists.txt combined is the culprit".

Thanks for the explanation, now I understand the problem better.

slurps-mad-rips:
(My apologies, I've had a string of personal life stuff get in the way of actually tackling the issues I had said I would. I'll be getting back to this PR later this week, hopefully 😞)

CMakeLists.txt
slurps-mad-rips: This has been removed as of the latest commit.

CMakeLists.txt
slurps-mad-rips: Let me know if the newer more broken up form is a bit easier to understand.

CMakeLists.txt
slurps-mad-rips: This turned out to be a typo on my part. The build succeeds in configuring (but fails tests otherwise it would seem 😅)

CMakeLists.txt
```diff
@@ -1,54 +1,105 @@
...
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-    if (NOT FOSSILIZE_ENABLE_EXCEPTIONS)
-        set(FOSSILIZE_EXCEPTION_FLAGS "-fno-exceptions")
-    else()
-        set(FOSSILIZE_EXCEPTION_FLAGS "")
-    endif()
-    set(FOSSILIZE_CXX_FLAGS -Wall -Wextra -Wunused-result -Wno-missing-field-initializers -Wno-empty-body -Wshadow ${FOSSILIZE_EXCEPTION_FLAGS})
+# option() cannot technically be used in newer CMake versions for non-boolean values
+# warnings will be thrown and this can be annoying otherwise.
+set(FOSSILIZE_VULKAN_INCLUDE_PATH "${PROJECT_SOURCE_DIR}/khronos/vulkan"
```
HansKristian-Work, line 45: Can this still be overriden by API users? I do that in another project.

HansKristian-Work:
Travis currently fails where it pulls in subproject tests:
https://travis-ci.org/github/ValveSoftware/Fossilize/jobs/685415054#L3604

I cannot reproduce this on master.

CMakeLists.txt
HansKristian-Work: It is, I can accept this. However, the -fvisibility=hidden is gone now.

CMakeLists.txt
slurps-mad-rips: Yes. If a project "above" Fossilize (i.e., fossilize is used via `add_subdirectory`) defines this variable, then this call will have no effect. Likewise, if you use `-DFOSSILIZE_VULKAN_INCLUDE_PATH=<path/to/vulkan>` when configuring CMake, this call will also have no effect :)

CMakeLists.txt
slurps-mad-rips: The visibillity is now set above on lines 23 and 27 respectively. CMake has a builtin variable to set this flag. 🙂

slurps-mad-rips:
> Travis currently fails where it pulls in subproject tests:
> https://travis-ci.org/github/ValveSoftware/Fossilize/jobs/685415054#L3604
> 
> I cannot reproduce this on master.
> 
> There is also a lot of weird spam about ".hidden" now:
> https://travis-ci.org/github/ValveSoftware/Fossilize/jobs/685415054#L1667

I'll try a few things on my end. I have an idea and it might involve some CMake global state that was fixed in a more recent version. 😅

slurps-mad-rips:
Apologies for not tending to this for a bit. Personal life has gotten in the way a bit 😬

I should be able to return to this soon™️

slurps-mad-rips:
Hi, apologies for never following up on this. I ended up realizing I would need to fix all of the CMake for most of the Khronos software stack to get this patch to work. On top of that I'm most likely not going to continue writing C++ professionally going forward, so I won't end up using Fossilize anyhow.

PR #83, opened 2020-05-19
Add headers for EXIT_ defines in CLI programs

aeikum:
The build failed on my Arch machine with gcc 10.1.0 and glibc 2.31. These macros are defined in stdlib.h, so let's include that: https://www.cplusplus.com/reference/cstdlib/


fooishbar:
Also fixes things for me on Fedora 32.

PR #94, opened 2020-09-04
RFC: Document StreamArchive file format

aeikum:
I know of a handful of implementations of this format, so we should formally define the format. A couple notes:

-The version was stored big-endian. Since the MS three bytes are currently zero, I've re-defined it to be one byte, with the upper three bytes unused. We could instead make it four big-endian bytes, if you like.

-I've defined the tag and hash sizes to be flexible. Together they must occupy 40 bytes, but the hash can be an application-defined length, with the tag taking up the remainder.

### DIFF
--- a/fossilize_db.cpp
+++ b/fossilize_db.cpp
@@ -610,26 +610,58 @@ DatabaseInterface *create_zip_archive_database(const char *path, DatabaseMode mo
 	return db;
 }
 
+/* Fossilize StreamArchive database format version 6:
+ *
+ * The file consists of a header, followed by an unlimited series of "entries".
+ *
+ * All multi-byte entities are little-endian.
+ *
+ * The file header is as follows:
+ *
+ * Field           Type           Description
+ * -----           ----           -----------
+ * magic_number    uint8_t[12]    Constant value: "\x81""FOSSILIZEDB"
+ * unused1         uint8_t        Currently unused. Must be zero.
+ * unused2         uint8_t        Currently unused. Must be zero.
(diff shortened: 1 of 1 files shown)


fossilize_db.cpp
```diff
@@ -610,27 +610,53 @@ DatabaseInterface *create_zip_archive_database(const char *path, DatabaseMode mo
...
+ * unused3         uint8_t        Currently unused. Must be zero.
+ * version         uint8_t        StreamArchive version: 6
+ *
+ *
+ * Each entry follows this format:
+ *
+ * Field           Type                              Description
+ * -----           ----                              -----------
+ * tag             unsigned char[40 - hash_bytes]    Application-defined 'tag' which groups entry types. Stored as hexadecimal big-endian ASCII.
+ * hash            unsigned char[hash_bytes]         Application-defined 'hash' to identify this entry. Stored as hexadecimal big-endian ASCII.
```
HansKristian-Work, line 635: Endianness of the hash is technically application defined. It should be treated as 40 opaque hex values.

HansKristian-Work (changes requested):
Sorry, didn't see this until now. Just a tiny nit, otherwise fine.

HansKristian-Work:
Another thing to mention is that the stream can be cut off anywhere and the last invalid entry should be considered to not exist.

aeikum:
Cool, good suggestions. I've pushed an update to fix those, and also mentions that tag+hash combos should be unique.

HansKristian-Work (approved):

PR #100, opened 2020-10-18
path: Do not use `getpid()` but use `/proc/self`

kakra:
The path `/proc/self` already references the path of the current pid.


kakra:
I think the Windows CI failures are unrelated to my changes.

HansKristian-Work:
CI failures are caused by outdated Python reference.

Issue #99, opened 2020-10-18
fossilize_replay should use batch scheduling if possible and share resources better


kakra:
When using `schedutil -B $(pgrep fossilize)` to put the replayer into batch scheduling mode, it utilizes my CPU better (99.9 to 100% per core instead of 91-98%), and it results in better IO rates writing the data back. This is probably because batch scheduling gives processes bigger time slices at the cost of latency but with better CPU cache hit rate in turn. This also results in better IO coalescing. Additionally, batch jobs should have a slight scheduling penalty over interactive processes, resulting in better system responsiveness while those are running.

But my biggest concern is with it's CPU weight and memory usage: Because those jobs use every core in my system, their CPU share is quite high above other processes. This even continues to work with a high CPU share while in games, and it seems to draw a lot of processing power from CPU and GPU even tho the processes run with nice=19.

So let's inspect this a little further: Modern distro kernels usually use auto-group CPU scheduling, especially when tailored for desktops. This makes sense, as CPU shares are distributed evenly among groups of processes instead of single processes: This essentially makes the scheduler see a complete application packages with all its sub-processes as a single process when it comes to distributing CPU shares. But this also means that nice=19 has absolutely no value because it is considered only among its own group of processes, the group itself will still receive its fair share - so a game running concurrently with the group of replay processes will only get 50% of the CPU. This has become very apparent to me since KDE switched to using slices and scopes for managing apps started from Plasma: I now see an app.slice which shares its total CPU bandwidth with other slices in the system, resulting in the share left to it evenly distributed between each scope running inside, with Steam and all its sub-processes being one such scope. I think Gnome deployed something similar. Maybe Steam just needs to take care that fossilize_replay doesn't create its own auto-group for the scheduler? I'm not if it does.

Also, when looking at memory accounting, I see the process group accounting for over 21 GB of memory on my system. The processes RSS itself only adds up to maybe roughly 8 GB. This is because memory accounting also accounts the page cache occupied by these processes. Or in other words: The replay jobs thrash the page cache, dominating it above other processes of the system. This slows loading times in games a lot.

[...]

PR #103, opened 2020-10-20
Autogroup batch scheduling

HansKristian-Work:
Improves the CPU scheduling situation on Linux.

Pulls in @kakra 's SCHED_BATCH enablement. Also, rather than use a setpgrp, use setsid to create a new session, which creates a new prgroup + makes a new autogroup for scheduling. This is apparently not strictly needed for purposes of SCHED_BATCH, but cannot hurt.

See discussion in #99.

### DIFF
--- a/fossilize_external_replayer_linux.hpp
+++ b/fossilize_external_replayer_linux.hpp
@@ -27,6 +27,7 @@
 #include <sys/wait.h>
 #include <fcntl.h>
 #include <unistd.h>
+#include <sched.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


kakra (approved):
LGTM. Is there a way to test this? I've found that operating the CLI fossilize uses a different code path than what Steam does. It looks like that code path is only used in external fossilize path, and that's probably only used when operated from the library.

HansKristian-Work:
Use --progress on the CLI. However, by default, the standalone CLI does not deal with inheriting process groups. This was originally done so that Ctrl+C would properly kill everything instead of daemonizing the process group, but that's fixed now. For now, you can set this to false: https://github.com/ValveSoftware/Fossilize/blob/master/cli/fossilize_replay.cpp#L2605. That should create a similar environment that Steam would use.

PR #105, opened 2020-10-22
Set Linux and OSX IO priority

kakra:
Sets the Linux IO priority in the main process to be inherited by its children later. For OSX I'm unsure about the scope of the IO policy we are setting but I'd expect it to be inherited by children.

@HansKristian-Work Broken out from #101


HansKristian-Work (approved):
LGTM.

PR #106, opened 2020-10-26
Rewrite how archives are parsed in a multi-process environment.

HansKristian-Work:
In order to stave off memory pressure and disk thrashing, we will parse
the archive once, and encode the resulting metadata in a shared memory
block. This metadata block can then be imported in child processes where
we avoid having to scan through the archive altogether. This saves a lot
of startup churn, and for very large archives + many processes, we save
a significant amount of memory as well, as essentially the same
duplicated information would be placed in memory.

### DIFF
--- a/cli/fossilize_replay.cpp
+++ b/cli/fossilize_replay.cpp
@@ -2436,7 +2436,8 @@ static void print_help()
 	"\t[--timeout <seconds>]\n" \
 	"\t[--progress]\n" \
 	"\t[--quiet-slave]\n" \
-	"\t[--shm-name <name>]\n\t[--shm-mutex-name <name>]\n"
+	"\t[--shm-name <name>]\n\t[--shm-mutex-name <name>]\n" \
+	"\t[--metadata-name <name>]\n"
 #else
 #define EXTRA_OPTIONS \
 	"\t[--slave-process]\n" \
@@ -2872,11 +2873,20 @@ static void populate_blob_hash_set(std::unordered_set<Hash> &hashes, ResourceTag
 		hashes.insert(h);
 }
 
-static int run_normal_process(ThreadedReplayer &replayer, const vector<const char *> &databases)
+static int run_normal_process(ThreadedReplayer &replayer, const vector<const char *> &databases, intptr_t metadata_handle)
 {
 	auto start_time = chrono::steady_clock::now();
 	auto start_create_archive = chrono::steady_clock::now();
 	auto resolver = create_database(databases);
+
+	if (DatabaseInterface::metadata_handle_is_valid(metadata_handle))
+	{
+		if (!resolver->import_metadata_from_os_handle(metadata_handle))
+		{
+			LOGE("Failed to import metadata.\n");
+			return EXIT_FAILURE;
+		}
+	}
 
 	auto end_create_archive = chrono::steady_clock::now();
 
@@ -3217,6 +3227,7 @@ int main(int argc, char *argv[])
 #ifdef _WIN32
 	const char *shm_name = nullptr;
 	const char *shm_mutex_name = nullptr;
+	const char *metadata_name = nullptr;
 #else
 	int shmem_fd = -1;
 #endif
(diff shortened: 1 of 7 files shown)


HansKristian-Work:
@kakra Here's a WIP PR for metadata block saving. Should save a lot of memory and helps a lot on startup time in general. Implemented the Linux side, gonna look at Windows side later.

kakra:
I think we should unmap all files from cache when all children are done. This can be done with `DONT_NEED`. But unlike the man page says, we should pass the file length to `fadvice()` - I didn't see kernel in the code that makes len=0 the complete file. It just happens that len=0 for RANDOM and SEQUENTIAL doesn't matter because it acts on the file handle instead of cached pages. It also triggers immediate write-back of outstanding dirty cache.

My latest patches in #101 are doing that currently whenever the database is closed - no matter if we are the parent or the child. That's not optimal as it will unmap the page cache also from the sibling processes and they don't stop all at the same time.

kakra:
I can probably test this later today in Linux. I may rebase my #101 onto this a refine the cache discarding code to work in the master process only.

HansKristian-Work:
I'm uncertain how meaningful it is to use fadvise beyond what we're doing in this PR (SEQUENTIAL vs RANDOM hints).

The only reasonable place we can emit DONTNEED for the replayer is after everything is complete. Is it really meaningful to call DONTNEED at this point? We're done processing, so there's no need why the system cannot reclaim those cached pages if needed.

Cannot DONTNEED for any writes should be okay, but considering that Fossilize does not write all that much to begin with on its own, I don't see a big win here.

HansKristian-Work:
I guess I can see one potential win when using DONTNEED at the very end, and that would somehow guide the kernel to start paging in any memory which was swapped out during replay.

kakra:
> I guess I can see one potential win when using DONTNEED at the very end

Yeah, that's my plan.

> and that would somehow guide the kernel to start paging in any memory which was swapped out during replay.

It's probably not working like that. It just prevents the kernel from swapping out any more memory from other processes because we told it that we no longer need our cache memory which will make the kernel free it. Swapping data back in that was possibly swapped out will happen on usage. Actually, on a system with enough memory, you'll see that it almost never swaps memory back in that was swapped out due to fossilize memory usage because that stale/inactive memory sitting around in some other process. But you will see that the kernel becomes encouraged by filling the freed memory with page cache for other applications again which is a win. Especially, a follow-up fossilize process has free memory to work with immediately resulting in possibly no additional swap out while it is running.

The remaining objective then would be to trigger immediate write-back of the driver shader cache so the kernel is able to immediately free that memory, too. According to the write-load of fossilize (PoE shaders) this holds around of 9 GB of cache for me which we "never" read again because fossilize uses a separate driver cache directory per game (at least when started from Steam).

My plan would be that we trick the driver into opening its cache files from the parent process, we could probably compile some dummy shader? Then, before tearing down the process, I'd scan `/proc/self/fd` for remaining open file descriptors and issue `fsync()` and DONTNEED on them. This should also work for mesa shaders as I expect it to open FDs on behalf of the process, too, and they should be in a private per-game directory, too. But I'm currently not able to verify or test this.

The final step would be to get amount of write-back data under control which should become a lot more easy once we have that previous objective in place.

I can imagine an additional hint we should know: If Steam is going to run fossilize as part of starting a game (that dialog saying "Wait while processing shader caches" or similar), we probably don't want to discard the driver shader cache from the page cache. I'm not sure if we can detect this with the current options but maybe by looking at the num-threads? It seems to be lower when Steam is running fossilize in the background while it runs on all cores when it'll run it in the foreground before starting a game.

HansKristian-Work:
> This should also work for mesa shaders as I expect it to open FDs on behalf of the process, too, and they should be in a private per-game directory, too. But I'm currently not able to verify or test this

I don't understand how this is going to work. The driver can open an FD, write the blob to disk, and then close the FD. There is no reasonable way Fossilize can do anything about that FD unless we inject debugger hooks or something.

kakra:
> I don't understand how this is going to work. The driver can open an FD, write the blob to disk, and then close the FD

At least the Linux NVIDIA driver keeps two FDs open for the life-time of the process. The FDs are opened on the first request that triggers shader compilation in the driver. I'm not sure how mesa does it. And Windows is probably a complete different story so I'm going to ignore that. We don't do cache memory discards for Windows currently anyways.

HansKristian-Work:
> I'm not sure how mesa does it.

Mesa doesn't do this, so you cannot assume driver specific implementation details like this.

HansKristian-Work:
I think the right approach for this is to fix things in drivers rather than trying to add questionable hacks like this. If there's a more global way to throttle IO writes, then I'd prefer that. The solution must be completely implementation agnostic.

kakra:
Yeah sure, that would be hacky but i'd see it more as a test if it has a potential to further improve things. If yes, it probably should indeed be fixed at the driver level. Otherwise I don't see a sane way to throttle the write-back rate. We could make heuristics by extrapolating the current load from loadavg but that is only effective if we can dynamically add and remove worker threads which the current design doesn't allow, I believe.

PR #118, opened 2021-01-06
Add support for dynamically adjusting load based on Linux PSI

HansKristian-Work:


kakra:
This looks cool... My kernel currently does not have PSI support but I probably can add that by recompiling. I will test with and without PSI.

HansKristian-Work:
@kakra This adds mechanisms for how we can dynamically adjust number of active processes. --log-memory will also report relevant information which can be useful. When there is IO/CPU/MEM pressure, the number of active processes is taken down. If there is lots of dirty pages being generated rapidly, the process count is also lowered.

cli/fossilize_replay_linux.hpp
```diff
@@ -476,6 +479,209 @@ bool ProcessProgress::start_child_process()
...
+	{
+		// Dirty pages is going down, can increase CPU load now.
+		go_up = true;
+	}
+
+	// There are other processes that would like some CPU time.
+	if (cpu_stall_ratio > 0.5f)
+		go_down = true;
+	else if (cpu_stall_ratio < 0.1f)
+		go_up = true;
```
kakra, line 606: Should we skip this check if fossilize runs in foreground mode? (that is when starting a game and Steam shows the progress dialog window for fossilize)

kakra (commented):
Looks like this could actually auto-tune itself to use just free resources...

cli/fossilize_replay_linux.hpp
kakra: BTW: After enabling PSI on my system, I'm pretty sure fossilize should not look at CPU pressure. In my system:
```
# cat /proc/pressure/cpu
some avg10=100.00 avg60=99.99 avg300=100.00 total=45537131575
```

This is because I have a very low priority daemon running that constantly uses at least one core.

So I suggest skipping this CPU pressure check completely and relying on the CPU scheduler instead (which we already switched to batch, with proper niceness of the process group). We could still use the `go_up` path of this check to add more threads when the CPU seems idle enough, tho.

kakra: However, it looks like `total` is not increasing at a 100% rate although `avg` suggests this... I'll look into that more deeply in the weekend.

kakra: Here's a graph of CPU pressure vs. CPU utilization vs. loadavg from one of our systems:

image
image
image

loadavg and CPU pressure seem to correlate very well.

This is IO pressure and memory pressure which also seem to correlate:

image
image

kakra: As we can see in the graphs, a stall on IO is almost always a full stall, so we should prioritize on keeping IO stall under control. Memory stall is less of a problem, it seems.

HansKristian-Work: > I think you can write a desired threshold to the file, then use `select/poll/epoll()`. That may be more accurate than manually calculating the values from diffs.

I tried this, but it requires root on my system since the /proc/pressure files are read-only for normal users, and we have to write to it to become a monitor. I think it's fine to do it like this. There is a timerfd that fires at regular intervals in the master process which does periodic polling anyways. We need to know the ratio, not just if a threshold was hit anyways.

HansKristian-Work: I agree on CPU stall factors being kinda meaningless.

I changed it to just consider IO related factors.

kakra: IO and memory may no longer correlate when the system is under memory pressure... OTOH, if swap is involved, IO will start to correlate again. In theory, this means we can get away with watching IO pressure only. In practice, I didn't test it and probably won't because my system was never affected in the way that other users reported (OOM killer kicking in etc).

kakra: > I tried this, but it requires root on my system

Ah okay... I also agree for the remainder of your answer.

kakra: > I agree on CPU stall factors being kinda meaningless.

Yeah, I think we should also stop using `nice` when Steam is running fossilize in foreground processing mode. `SCHED_BATCH` is still useful, tho (because better throughput using longer scheduler timeslices).

kakra: Is it possible for fossilize to detect foreground vs background mode?

HansKristian-Work: I'm considering adding a way for the controlling process to dynamically enable/disable PSI usage. Not sure if we should bother with nice. Since we now have a setup to control number of active processes dynamically it should be easy to say something like "only keep N processes active in background, but when going to foreground, light up all cores", something like that.

kakra: > dynamically enable/disable PSI usage

I think PSI usage should be watched the whole time, no matter if we a limited to all cores or just a reduced number of cores (background mode). The original issue report was about fossilize overwhelming the available system resources - and watching IO pressure is the only active thing we can do to prevent that situation. While the user waits for a foreground fossilize to finish, they still might want to use the system for other tasks. Similarly, watching the dirty cache ratio is still useful but may be relaxed in foreground mode.

BTW: Is it possible to just replace the fossilize binary deployed by Steam with a compiled version from this branch for testing? OTOH, it won't run anyways currently because I'm on NVIDIA. Is there a way to override that behavior?

HansKristian-Work: It's possible to replace the fossilize replayer binary I think. However, the external launcher code is linked into Steam I think, so that's harder to change. I don't know much if anything about the Steam side, I just work on this repo.

HansKristian-Work: Ok, made it so that the external replayer can send messages to the replayer about options. It can then force number of active processes, enable if IO stalling matters, etc. That should enable Steam to dynamically change between background/foreground type scheduling.

Going beyond this will be extreme overkill, so I'll probably leave it here. Need to test a bit more later.

kakra: > However, the external launcher code is linked into Steam I think

Ah okay yes, that's problematic then. This probably also means it cannot be distributed as a separate package which could be switched to beta (like Proton).

> That should enable Steam to dynamically change between background/foreground type scheduling.

I think that's more than enough. I just asked if it would be possible. ;-)

kakra (approved):
This seems to work pretty great: While loading the system with a Steam job to verify game files, fossilize adjusts itself and stays at around 20-120 MB dirty, and IO PSI stays below 50%. But fossilize fails to make significant progress at times (with no workers running for short periods).

@HansKristian-Work I think this behavior is expected but I'm thinking about your original idea of not forcing fossilize to not make any progress when you reviewed my loadavg based proof of concept. But nevetheless: Even when it generates a loadavg of 25 (and while Steam verifies data, I've seen peaks of 190), the system behaves much better and I'm seeing no desktop stalls, browsing the web in foreground works without problems.

While fossilize is running without any concurrent activity, it mostly stays at 8 workers (on 8 cores) until the dirty background flusher of the kernel kicks in, it then drops to 4-6 workers for a few seconds, then goes back to 8 workers. So this is working very well it seems.

Well done! :+1: :1st_place_medal:

kakra:
BTW: I think with this in place, and the scheduling updates, Steam may even start fossilize in background processing mode with more than just two workers: Since fossilize is now careful about writeback fluctuations, it should no longer dominate the cache by flooding it with dirty data.

kakra:
For the rest of the issues (amount of write-back generated) it's probably NVIDIAs job to inspect why the driver is generating so much written data even for mostly clean caches.

PR #131, opened 2021-05-06
Fix some process leaks.

### LINKED ISSUE #129: Steam leaks fossilize_replay children in stopped state
When Steam crashes (and maybe also when quit normally) while fossilze_replay is running, it may leak fossilize_replay processes in stopped state (as part of the load limiter which `SIG_STOP`s children). The main process should probably send `SIG_CONT` to all children before exiting or being forcefully quit.

HansKristian-Work:
- Need to use SIGKILL to ensure that stopped child processes actually
  die. The replayer handled this correctly, but not the external
  replayer API.
- Need to use SEQPACKET socket type to actually be able to detect
  hangups. socketpair() DGRAMs apparently don't work even if they are
  connected ...

Fix #129.

### DIFF
--- a/fossilize_external_replayer_linux.hpp
+++ b/fossilize_external_replayer_linux.hpp
@@ -418,7 +418,7 @@ bool ExternalReplayer::Impl::kill()
 		kill_fd = -1;
 	}
 
-	bool ret = killpg(pid, SIGTERM) == 0;
+	bool ret = killpg(pid, SIGKILL) == 0;
 	if (!ret)
 		LOGI("ExternalReplayer::Impl::kill(): Failed to kill: errno %d.\n", errno);
 	return ret;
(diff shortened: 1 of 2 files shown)


fossilize_external_replayer_linux.hpp
```diff
@@ -418,7 +418,7 @@ bool ExternalReplayer::Impl::kill()
 		kill_fd = -1;
 	}
 
-	bool ret = killpg(pid, SIGTERM) == 0;
+	bool ret = killpg(pid, SIGKILL) == 0;
```
kakra, line 421: I'm not sure if this is enough because the leaking processes in stopped state didn't go away when I sent SIGKILL manually, it needed an additional SIGCONT. But let's see how it changes once this update is deployed.

HansKristian-Work: I tested this and SIGKILLing a stopped child worked just fine.

PR #133, opened 2021-05-22
Close unused crash files after forking

olvaffe:
When a child process is forked, it keeps the crash files of all prior
child processes open.  We need to close them, otherwise

 - when a child process dies and its crash_handler is not invoked, and
   if its crash file is kept open by other child processes, the master
   process's call to epoll_wait will return immediately with an EPOLLHUP
   event.  Its main loop becomes a busy loop.
 - same as above, but with the crash_handler invoked, the master process
   will respawn the child process.  epoll_wait will still return
   immediately with an EPOLLHUP event.  But since the ProcessProgress
   has been updated to track the respawned child process, fgets called
   from process_once will block forever until the respawned child
   process dies/exits.

This is tested with "fossilize-replay --master-process --num-threads 3",
and killing the child processes with "kill -TERM" or "kill -ABRT"

### DIFF
--- a/cli/fossilize_replay_linux.hpp
+++ b/cli/fossilize_replay_linux.hpp
@@ -101,7 +101,7 @@ struct ProcessProgress
 
 	bool process_once();
 	bool process_shutdown(int wstatus);
-	bool start_child_process();
+	bool start_child_process(vector<ProcessProgress> &siblings);
 	void parse(const char *cmd);
 
 	uint32_t index = 0;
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Nice find! Weird that I never ran into this when testing this ... I also added explicit EPOLL_CTL_DEL to avoid any such issue later.

olvaffe:
I could always reproduce the issues during my testing, except for when I killed the last child process.  That might explain it?

PR #168, opened 2022-01-28
Add support for few missing pipeline and sampler vk structs

hakzsam:
This PR adds support for the following missing vk structs:

- VkPipelineColorWriteCreateInfoEXT
- VkPipelineRasterizationProvokingVertexStateCreateInfoEXT
- VkSamplerCustomBorderColorCreateInfoEXT
- VkSamplerReductionModeCreateInfo


fossilize.cpp
```diff
@@ -3493,6 +3559,59 @@ bool StateReplayer::Impl::parse_pipeline_rendering_info(const Value &state,
...
+	return true;
+}
+
+bool StateReplayer::Impl::parse_sampler_custom_border_color(const Value &state,
+							    VkSamplerCustomBorderColorCreateInfoEXT **out_info)
+{
+	auto *info = allocator.allocate_cleared<VkSamplerCustomBorderColorCreateInfoEXT>();
+	*out_info = info;
+
+	info->customBorderColor.uint32[0] = state["customBorderColor"]["0"].GetUint();
```
HansKristian-Work, line 3595: This should be an array in JSON.

fossilize.cpp
```diff
@@ -6291,6 +6506,64 @@ static bool json_value(const VkPipelineRenderingCreateInfoKHR &create_info, Allo
...
+}
+
+template <typename Allocator>
+static bool json_value(const VkSamplerCustomBorderColorCreateInfoEXT &create_info, Allocator &alloc, Value *out_value)
+{
+	Value value(kObjectType);
+	value.AddMember("sType", create_info.sType, alloc);
+
+	Value customBorderColor(kObjectType);
+	customBorderColor.AddMember("0", create_info.customBorderColor.uint32[0], alloc);
```
HansKristian-Work, line 6545: Should be kArrayType.

fossilize.cpp
```diff
@@ -4701,10 +4908,12 @@ bool StateRecorder::Impl::copy_shader_module(const VkShaderModuleCreateInfo *cre
 bool StateRecorder::Impl::copy_sampler(const VkSamplerCreateInfo *create_info, ScratchAllocator &alloc,
                                        VkSamplerCreateInfo **out_create_info)
 {
-	if (create_info->pNext)
+	auto *info = copy(create_info, 1, alloc);
+
+	if (!copy_pnext_chain(info->pNext, alloc, &info->pNext))
```
HansKristian-Work, line 4913: I think we're missing pNext hashing of samplers?

fossilize.cpp
```diff
@@ -6291,6 +6506,64 @@ static bool json_value(const VkPipelineRenderingCreateInfoKHR &create_info, Allo
 	return true;
 }
 
+template <typename Allocator>
+static bool json_value(const VkPipelineColorWriteCreateInfoEXT &create_info, Allocator &alloc, Value *out_value)
```
HansKristian-Work, line 6510: One thing about ColorWriteCreateInfo in particular is that it is affected by dynamic state, e.g. VK_DYNAMIC_STATE_COLOR_WRITE_ENABLE_BIT. I think we have to ignore color write enables array when hashing.

HansKristian-Work (changes requested):
Overall seems okay, mostly missing roundtrip tests for these.

fossilize.cpp
hakzsam: Yeah, I think the main problem is that we can't access the dynamic_info struct from there currently?

fossilize.cpp
hakzsam: Fixed.

fossilize.cpp
hakzsam: Should be fixed now.

fossilize.cpp
hakzsam: Fixed.

fossilize.cpp
```diff
@@ -4363,11 +4576,18 @@ void StateRecorder::Impl::pump_synchronized_recording(StateRecorder *recorder)
 bool StateRecorder::record_sampler(VkSampler sampler, const VkSamplerCreateInfo &create_info, Hash custom_hash)
 {
 	{
-		if (create_info.pNext)
+		// Only custom border color and reduction mode are supported for now.
+		auto *custom_border_color = find_pnext<VkSamplerCustomBorderColorCreateInfoEXT>(
```
HansKristian-Work, line 4580: This is not necessary to check explicitly. We'll just error out in copy_pnext_chain instead.

fossilize.cpp
HansKristian-Work: Just add a TODO, and I think we can let it slide for now.

hakzsam: Added a TODO.

HansKristian-Work (approved):
LGTM.

PR #175, opened 2022-02-01
Fix serialization of immutable samplers

hakzsam:


fossilize.cpp
```diff
@@ -7402,7 +7402,8 @@ static bool json_value(const VkDescriptorSetLayoutCreateInfo& layout, Allocator&
 		binding.AddMember("descriptorCount", b.descriptorCount, alloc);
 		binding.AddMember("stageFlags", b.stageFlags, alloc);
 		binding.AddMember("binding", b.binding, alloc);
-		if (b.pImmutableSamplers)
+		if (b.pImmutableSamplers && b.descriptorCount > 0 &&
+		    (b.descriptorType == VK_DESCRIPTOR_TYPE_SAMPLER || b.descriptorType == VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER))
```
HansKristian-Work, line 7406: It makes more sense to fix copy_descriptor_set_layout() instead. There, we forget to clear pImmutableSamplers to nullptr if we hit this scenario.

hakzsam: Yeah, it's much better.

HansKristian-Work (approved):
LGTM

PR #177, opened 2022-02-01
Fix serialization of dynamic blend constants

hakzsam:


fossilize.cpp
```diff
@@ -8041,7 +8041,7 @@ static bool json_value(const VkGraphicsPipelineCreateInfo &pipe,
 		cb.AddMember("logicOpEnable", pipe.pColorBlendState->logicOpEnable, alloc);
 		Value blend_constants(kArrayType);
 		for (auto &c : pipe.pColorBlendState->blendConstants)
-			blend_constants.PushBack(c, alloc);
+			blend_constants.PushBack(dynamic_info.blend_constants ? 0 : c, alloc);
```
HansKristian-Work, line 8044: `0.0f` is preferable. I assume the int will be auto-promoted to float, but it gives me pause.

hakzsam: Yes, it's more correct that way.

HansKristian-Work (approved):
LGTM

PR #180, opened 2022-02-01
Fix missing validation of mutable samplers

hakzsam:


cli/fossilize_feature_filter.cpp
```diff
@@ -1000,6 +1003,11 @@ bool FeatureFilter::Impl::descriptor_set_layout_is_supported(const VkDescriptorS
 					mutable_counts.storage_image = 1;
 					break;
 
+				case VK_DESCRIPTOR_TYPE_SAMPLER:
+					// Sampler is a non-standard type for mutable
+					must_check_set_layout_before_accept = true;
```
HansKristian-Work, line 1008: Should also set the mutable counts I think. It just also needs to set this bool to true.

cli/device.cpp
```diff
@@ -518,6 +518,17 @@ bool VulkanDevice::format_is_supported(VkFormat format, VkFormatFeatureFlags for
...
 }
 
+bool VulkanDevice::descriptor_set_layout_is_supported(const VkDescriptorSetLayoutCreateInfo *info)
+{
+	if (is_null_device)
+		return true;
+
+	VkDescriptorSetLayoutSupport set_layout_support = { VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_SUPPORT };
+
+	vkGetDescriptorSetLayoutSupport(device, info, &set_layout_support);
```
HansKristian-Work, line 528: Probably a little more safe to use the KHR one in case a 1.0 device is used. maintenance3 is required for mutable, not 1.1.

hakzsam: Sounds good.

cli/fossilize_feature_filter.cpp
hakzsam: Should be fixed.

HansKristian-Work (approved):
LGTM

PR #181, opened 2022-02-03
Fix hashing color blend state if blend isn't enabled

hakzsam:


HansKristian-Work (approved):
LGTM, derpy oversight :\

PR #183, opened 2022-02-10
Add request_shutdown() which will terminate prepare().  This can be u…

danginsburg:
…sed if prepare() is running on a thread and the thread needs to shutdown while it is running (for example, on very large archives).

### DIFF
--- a/fossilize_db.cpp
+++ b/fossilize_db.cpp
@@ -119,6 +119,9 @@ struct ExportedMetadataHeader
 };
 static_assert(sizeof(ExportedMetadataHeader) % 8 == 0, "Alignment of ExportedMetadataHeader must be 8.");
 
+// Allow termination request if using the interface on a thread
+std::atomic<bool> shutdown_requested;
+
 struct DatabaseInterface::Impl
 {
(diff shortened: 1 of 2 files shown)


fossilize_db.cpp
```diff
@@ -119,6 +119,9 @@ struct ExportedMetadataHeader
 };
 static_assert(sizeof(ExportedMetadataHeader) % 8 == 0, "Alignment of ExportedMetadataHeader must be 8.");
 
+// Allow termination request if using the interface on a thread
+std::atomic<bool> shutdown_requested;
```
HansKristian-Work, line 123: Is there a need for this to be a static global? I would imagine something like `database_iface->request_shutdown()` would work too.

danginsburg: That was my first implementation, but when the extra read only database was used it wouldn't terminate those.  I would have to connect those somehow to the other database I guess?

HansKristian-Work: Right ye ... Could get spicy w.r.t. thread safety as well.

HansKristian-Work (approved):

PR #187, opened 2022-03-18
Serialize VRS features and add a workaround for vkd3d

hakzsam:
RADV enables a shader workaround due to a hardware bug based on these features.

### DIFF
--- a/cli/device.cpp
+++ b/cli/device.cpp
@@ -153,6 +153,17 @@ static bool application_info_promote_robustness2(const VkApplicationInfo *app_in
 	return false;
 }
 
+static bool application_info_promote_fragment_shading_rate(const VkApplicationInfo *app_info)
+{
+	if (!app_info || !app_info->pEngineName)
+		return false;
+
+	if (strcmp("vkd3d", app_info->pEngineName) == 0)
+		return true;
+
+	return false;
+}
+
 bool VulkanDevice::init_device(const Options &opts)
 {
 	if (opts.null_device)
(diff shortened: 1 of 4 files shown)


cli/device.cpp
```diff
@@ -368,20 +379,41 @@ bool VulkanDevice::init_device(const Options &opts)
...
+			requested_pdf2 = &replacement_pdf2;
+		}
+
+		if (application_info_promote_fragment_shading_rate(opts.application_info) &&
+		    find_pnext<VkPhysicalDeviceFragmentShadingRateFeaturesKHR>(
+				    VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAGMENT_SHADING_RATE_FEATURES_KHR,
+				    opts.features->pNext) == nullptr)
+		{
+			replacement_fragment_shading_rate = { VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAGMENT_SHADING_RATE_FEATURES_KHR };
+			replacement_pdf2 = *requested_pdf2;
```
HansKristian-Work, line 408: The branch above copies PDF2 data from opts.features. Probably a good idea to harmonize the two.

cli/device.cpp
```diff
@@ -368,20 +379,41 @@ bool VulkanDevice::init_device(const Options &opts)
...
+			replacement_pdf2.pNext = &replacement_fragment_shading_rate;
+			replacement_fragment_shading_rate.pNext = requested_pdf2->pNext;
```
HansKristian-Work, line 410: If replacement_pdf2.pNext is written here. Wouldn't that potentially disturb requested_pdf2 as well? We probably need to write replacement_fragment_shading_rate->pNext, then the PDF2.

cli/device.cpp
hakzsam: Sounds good.

cli/device.cpp
hakzsam: Should be fixed now.

PR #190, opened 2022-03-18
replay: Make directory-based DB non-fatal

ishitatsuyuki:
We should not refuse to load directory-based DBs (e.g. generated by
RenderDoc) just because we can't calculate compression statistics.


HansKristian-Work (approved):
LGTM

Issue #213, opened 2023-01-25
dynamic state3 + GPL brokenness


zmike:
A couple different items:

Zink uses GPL with a fragment output stage that uses the forbidden trio of `VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT`,
`VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT`, and `VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT`. According to the spec:

`
pAttachments is a pointer to an array of VkPipelineColorBlendAttachmentState structures defining blend state for each color attachment. It is ignored if the pipeline is created with VK_DYNAMIC_STATE_COLOR_BLEND_ENABLE_EXT, VK_DYNAMIC_STATE_COLOR_BLEND_EQUATION_EXT, and VK_DYNAMIC_STATE_COLOR_WRITE_MASK_EXT dynamic states set.
`

This is not handled correctly, and fossilize crashes from attempting to access the null `pAttachments` pointer.

Additionally, `VkPipelineRenderingCreateInfo` is accessed improperly in non-fragment_output stages for GPL pipelines. According to spec, formats may only be accessed in fragment output stage.

PR #214, opened 2023-01-25
GPL + EDS3 fixes

zmike:
this should resolve https://github.com/ValveSoftware/Fossilize/issues/213


zmike:
Hm it seems like it might not be correctly replaying the captures, so probably I missed/broke something else too.

PR #215, opened 2023-01-26
Misc EDS3 and GPL serialization fixes.

### LINKED ISSUE #213: dynamic state3 + GPL brokenness

HansKristian-Work:
Fix #213.
Supersedes #214.

### DIFF
--- a/fossilize.cpp
+++ b/fossilize.cpp
@@ -1967,7 +1967,20 @@ bool compute_hash_graphics_pipeline(const StateRecorder &recorder, const VkGraph
 
 		bool need_blend_constants = false;
 
-		for (uint32_t i = 0; i < b.attachmentCount; i++)
+		// Special EDS3 rule. If all of these are set, we must ignore the pAttachments pointer.
+		const bool dynamic_attachments = dynamic_info.color_blend_enable &&
+		                                 dynamic_info.color_write_mask &&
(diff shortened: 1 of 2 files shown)


zmike:
Works!

PR #222, opened 2023-04-27
Explicitly include stdint header

cmarcelo:
Fix build for g++-13.

### DIFF
--- a/path.hpp
+++ b/path.hpp
@@ -21,6 +21,7 @@
  */
 
 #pragma once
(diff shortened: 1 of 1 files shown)


path.hpp
```diff
@@ -21,6 +21,7 @@
  */
 
 #pragma once
+#include <cstdint>
```
HansKristian-Work, line 24: stdint.h, not cstdint.

PR #245, opened 2024-05-15
Implement nested pipeline libraries

KonstantinSeurer:


cli/fossilize_replay.cpp
```diff
@@ -2467,6 +2447,60 @@ struct ThreadedReplayer : StateCreatorInterface
...
 	}
 
+	template <typename DerivedInfo>
+	uint32_t pipeline_library_get_parent_depth(const VkPipelineLibraryCreateInfoKHR *library,
+	                                           const vector<DerivedInfo> *deferred,
+	                                           const unordered_map<Hash, DerivedInfo> &parents)
+	{
+		uint32_t max_depth = 0;
+
+		for (uint32_t i = 0; i < library->libraryCount; i++)
```
HansKristian-Work, line 2457: I don't think there is even any guarantee here that a derived pipeline has been parsed, so I don't understand how you can compute the parent depth.

HansKristian-Work (changes requested):

cli/fossilize_replay.cpp
```diff
@@ -2163,17 +2153,8 @@ struct ThreadedReplayer : StateCreatorInterface
 		if (opts.pipeline_stats)
 			info->flags |= VK_PIPELINE_CREATE_CAPTURE_STATISTICS_BIT_KHR;
 
-		auto *library_info = find_pnext<VkPipelineLibraryCreateInfoKHR>(
-				VK_STRUCTURE_TYPE_PIPELINE_LIBRARY_CREATE_INFO_KHR, create_info->pNext);
-		bool consumes_libraries = library_info && library_info->libraryCount != 0;
 		bool generates_library = (create_info->flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR) != 0;
 
-		if (generates_library && consumes_libraries)
```
HansKristian-Work, line 2171: I only see code that attempts to do this for RT.

cli/fossilize_replay.cpp
```diff
@@ -2104,18 +2104,8 @@ struct ThreadedReplayer : StateCreatorInterface
 		if (opts.pipeline_stats)
 			info->flags |= VK_PIPELINE_CREATE_CAPTURE_STATISTICS_BIT_KHR;
 
-		// Compute does not support pipeline libraries, but sanitizes the code paths.
-		auto *library_info = find_pnext<VkPipelineLibraryCreateInfoKHR>(
-				VK_STRUCTURE_TYPE_PIPELINE_LIBRARY_CREATE_INFO_KHR, create_info->pNext);
-		bool consumes_libraries = library_info && library_info->libraryCount != 0;
 		bool generates_library = (create_info->flags & VK_PIPELINE_CREATE_LIBRARY_BIT_KHR) != 0;
 
-		if (generates_library && consumes_libraries)
```
HansKristian-Work, line 2113: I only see code that attempts to do this for RT.

cli/fossilize_replay.cpp
KonstantinSeurer: `work_item_get_parent_depth` should handle all pipeline types. Are there other code paths that miss an implementation?

HansKristian-Work: You're not setting hash_to_index[hash] for those at least. In the code you're looking at the create info to determine the depth, but how can you know the create info when the pipeline has not even been parsed yet?

KonstantinSeurer: I added the missing hash_to_index (hash_to_info now) handling. Parsing pipeline libraries should be fixed as well.

cli/fossilize_replay.cpp
```diff
@@ -2822,27 +2893,45 @@ struct ThreadedReplayer : StateCreatorInterface
...
+						                progress = false;
+
+						                // Wait until all parent pipelines have been parsed.
+						                sync_worker_memory_context(PARENT_PIPELINE_MEMORY_CONTEXT);
+
+						                // Queue up all shader modules if somehow the shader modules used by parent pipelines differ from children ...
+						                for (auto &parent : parents)
+						                {
+							                // Handle nested libraries.
+							                progress |= enqueue_parent_pipelines(parent.second, pipelines, *parsed_parents);
```
HansKristian-Work, line 2920: This is a race condition since enqueue_parent_pipelines will push work that ends up writing to `parents`, and you're iterating over it here.

cli/fossilize_replay.cpp
```diff
@@ -2685,6 +2751,11 @@ struct ThreadedReplayer : StateCreatorInterface
 					                 }
 				                 }
 
+				                 for (const auto& item : deferred[memory_index])
+				                 {
+					                 hash_to_info.erase(item.hash);
```
HansKristian-Work, line 2756: This is a race condition I think? Erasing stuff while parsing happens.

KonstantinSeurer:
As you suggested, the depth is now computed by enqueue_parent_pipelines. enqueue_parent_pipelines now writes the hashes of enqueued pipelines to a set of which the parent pipelines are enqueued until all dependencies are resolved. The loop over max_depth has been replaced by a vector which is sorted by compilation order.

cli/fossilize_replay.cpp
```diff
@@ -2575,47 +2567,62 @@ struct ThreadedReplayer : StateCreatorInterface
 
 		if (is_outside_range)
 		{
+			if (parents.count(h) != 0) {
+				auto &parent = parents.at(h);
+				parent.depth = std::max(parent.depth, depth);
```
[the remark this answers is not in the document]
HansKristian-Work: Seeing that depth can spuriously increase later due to a difference dependency chain requiring a pipeline, I don't think trying to resolve depth here is appropriate after all.

Maybe the correct thing to do here is to parse parents to fill out `parents`, ignore depth, then keep parsing until there are no new dependees in flight. Then, to a recursive depth-first traversal to fill in depth and write out a linear array at the same time, similar to a post-depth traversal CFG.

cli/fossilize_replay.cpp
```diff
@@ -2660,6 +2667,8 @@ struct ThreadedReplayer : StateCreatorInterface
 			       memory_index;
 		};
 
+		auto parsed_parents = make_shared<unordered_set<Hash>>();
```
HansKristian-Work, line 2670: Is this really needed?

cli/fossilize_replay.cpp
```diff
@@ -2776,6 +2785,8 @@ struct ThreadedReplayer : StateCreatorInterface
 									 // if we keep them around indefinitely.
 					                 free_pipelines();
 
+					                 parents.clear();
```
HansKristian-Work, line 2788: Isn't this cleared elsewhere already?

cli/fossilize_replay.cpp
```diff
@@ -2822,20 +2833,36 @@ struct ThreadedReplayer : StateCreatorInterface
...
+				                [this, &pipelines, derived, &parents, parsed_parents, hash_offset, start_index]()
 				                {
-					                // Wait until all parent pipelines have been parsed.
-					                sync_worker_memory_context(PARENT_PIPELINE_MEMORY_CONTEXT);
+					                while (true)
+					                {
+						                // Wait until all parent pipelines have been parsed.
+						                sync_worker_memory_context(PARENT_PIPELINE_MEMORY_CONTEXT);
+
+						                unordered_set<Hash> dependencies = *parsed_parents;
```
HansKristian-Work, line 2853: Would it make more sense to leave `dependencies` outside the loop and do `dependencies.clear(); std::swap(*parsed_parents, dependencies)` instead? Should avoid the deep copy.

cli/fossilize_replay.cpp
```diff
@@ -2866,36 +2893,50 @@ struct ThreadedReplayer : StateCreatorInterface
...
+					                sort(begin(ordered_parents), end(ordered_parents), [&](const DerivedInfo &a, const DerivedInfo &b) -> bool {
+						                return a.depth > b.depth;
+					                });
 
 					                // We might be pulling in a parent pipeline from another memory context next iteration,
 					                // so we need to wait for all normal memory contexts.
 					                for (unsigned i = 0; i < NUM_PIPELINE_MEMORY_CONTEXTS; i++)
 						                sync_worker_memory_context(i);
+
+					                uint32_t prev_depth = 0xFFFFFFFF;
```
HansKristian-Work, line 2910: UINT32_MAX?

cli/fossilize_replay.cpp
```diff
@@ -2866,36 +2893,50 @@ struct ThreadedReplayer : StateCreatorInterface
...
+					                for (auto &parent : ordered_parents)
+					                {
+						                if (!parent.info)
+							                continue;
+
+						                if (!derived_work_item_is_satisfied(parent, pipelines))
+							                continue;
+
+						                if (prev_depth != parent.depth) {
```
HansKristian-Work, line 2919: Brace on next line.

cli/fossilize_replay.cpp
```diff
@@ -2866,36 +2893,50 @@ struct ThreadedReplayer : StateCreatorInterface
...
 			}
 
 			work.push_back({ get_order_index(ENQUEUE_DERIVED_PIPELINES_OFFSET),
-			                 [this, &pipelines, derived, memory_index, hash_offset, start_index]() {
+			                 [this, &pipelines, derived, &parents, memory_index, hash_offset, start_index]() {
 				                 // Go over all pipelines. If there are no further dependencies to resolve, we can go ahead and queue them up.
 				                 // If an entry exists in pipelines, we have queued up that hash earlier, but it might not be done compiling yet.
 				                 auto itr = unstable_remove_if(begin(*derived), end(*derived), [&](const DerivedInfo &info) -> bool {
-					                 return derived_work_item_is_satisfied(info, pipelines);
+					                 return derived_work_item_is_satisfied(info, pipelines) || parents.count(info.hash) != 0;
```
HansKristian-Work, line 2939: I don't understand this line.

cli/fossilize_replay.cpp
```diff
@@ -2913,7 +2954,7 @@ struct ThreadedReplayer : StateCreatorInterface
 				                 // Now we can enqueue pipeline compilation with correct pipeline handles.
 				                 for (auto i = itr; i != end(*derived); ++i)
 				                 {
-					                 if (i->info)
+					                 if (i->info && parents.count(i->hash) == 0)
```
HansKristian-Work, line 2957: Or this one.

cli/fossilize_replay.cpp
KonstantinSeurer: Nested pipeline libraries are already compiled in ENQUEUE_SHADER_MODULE_SECONDARY_OFFSET. This and the change below filter out pipelines that were already compiled without triggering the `pipelines were not compiled because parent pipelines do not exist.` warning. I can add a comment or change the condition for the warning to make it clear what's happening.

cli/fossilize_replay.cpp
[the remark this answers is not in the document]
KonstantinSeurer: Sounds good.

cli/fossilize_replay.cpp
KonstantinSeurer: I moved it because it's needed in ENQUEUE_DERIVED_PIPELINES_OFFSET.

cli/fossilize_replay.cpp
KonstantinSeurer: I moved it here because `parents` is used by `ENQUEUE_DERIVED_PIPELINES_OFFSET`.

cli/fossilize_replay.cpp
KonstantinSeurer: It's needed to avoid calling `enqueue_parent_pipelines` for pipelines that were already parsed.

cli/fossilize_replay.cpp
```diff
@@ -2554,38 +2546,54 @@ struct ThreadedReplayer : StateCreatorInterface
...
 		bool is_outside_range = pipelines.count(h) == 0;
 
-		if (is_outside_range)
+		if (is_outside_range && parsed_parents.count(h) == 0)
 		{
+			{
+				lock_guard<mutex> holder(hash_lock);
+				if (parents.count(h) != 0) {
+					auto &parent = parents.at(h);
+					parent.depth = std::max(parent.depth, depth);
```
HansKristian-Work, line 2555: I still don't think this works, because if parent has dependencies, that chain of dependencies must also be moved up in depth accordingly.

I'm not sure we can avoid doing the traversal after all pipelines have been parsed.

cli/fossilize_replay.cpp
```diff
@@ -2490,47 +2496,22 @@ struct ThreadedReplayer : StateCreatorInterface
...
-		return derived_work_item_is_satisfied_base(info, pipelines);
-	}
-
-	bool derived_work_item_is_satisfied(const DeferredRayTracingInfo &info,
 	                                    const unordered_map<Hash, VkPipeline> &pipelines) const
 	{
-		if (!derived_work_item_is_satisfied_base(info, pipelines))
-			return false;
-		return !info.info->pLibraryInfo || pipeline_library_info_is_satisfied(*info.info->pLibraryInfo, pipelines);
+		auto *library = work_item_get_library_info(info);
```
HansKristian-Work, line 2493: This loses the condition that if info.info is nullptr (from work_item_is_derived), false is returned. It will instead return true here.

cli/fossilize_replay.cpp
HansKristian-Work: Also, brace on next line.

HansKristian-Work: For example, consider A (depth 1) -> B (depth 2) -> C (depth 3).
If D -> E -> F -> B, B needs to be adjusted to depth 4, but it also means that C must be updated to depth 5. Doing it recursively inline is a potential exponential explosion in complexity, so that's not really an option I think.

HansKristian-Work (changes requested):
I think the first commit can be a PR on its own with the fix applied. Anything to reduce incremental complexity in this PR is good.

cli/fossilize_replay.cpp
KonstantinSeurer: `work_item_get_library_info` already returns `nullptr` if `info.info` is `nullptr`.

HansKristian-Work: Yes, it does, but then you check if `!library`, which becomes true.

cli/fossilize_replay.cpp
KonstantinSeurer: Right, not sure how I expected this to work in the first place. I have a local fixup that parses all parents, gathers all top level ones and recursively computes their patents' depth.

cli/fossilize_replay.cpp
KonstantinSeurer: oh

cli/fossilize_replay.cpp
```diff
@@ -2568,36 +2550,61 @@ struct ThreadedReplayer : StateCreatorInterface
...
+		}
 	}
 
 	template <typename DerivedInfo>
 	void enqueue_deferred_pipelines(vector<DerivedInfo> *deferred, const unordered_map<Hash, VkPipeline> &pipelines,
 	                                unordered_map<Hash, DerivedInfo> &parents,
 	                                vector<EnqueuedWork> &work, const vector<Hash> &hashes, unsigned start_index)
 	{
-		static const unsigned NUM_PIPELINES_PER_CONTEXT = 1024;
+		static const unsigned NUM_PIPELINES_PER_CONTEXT = 1;
```
HansKristian-Work, line 2607: ?

cli/fossilize_replay.cpp
```diff
@@ -2568,28 +2550,53 @@ struct ThreadedReplayer : StateCreatorInterface
...
+	}
+
+	template <typename DerivedInfo>
+	void compute_parents_depth(const unordered_map<Hash, DerivedInfo> &parents,
+	                           unordered_map<Hash, uint32_t> &parents_depth,
+	                           const DerivedInfo &info, uint32_t depth)
+	{
+		parents_depth[info.hash] = max(parents_depth.at(info.hash), depth);
+
+		auto library = work_item_get_library_info(info);
```
HansKristian-Work, line 2590: This leads to exponential explosion as well. I think what can be done here is to compare depth against parents depth. Only if depth is greater than parents_depth does the recursive descend happen. That way we can terminate the traversal early.

cli/fossilize_replay.cpp
```diff
@@ -2796,20 +2807,38 @@ struct ThreadedReplayer : StateCreatorInterface
...
+					                unordered_map<Hash, DerivedInfo> top_level_parents = parents;
+
+					                unordered_set<Hash> dependencies;
+					                while (true)
+					                {
+						                // Wait until all parent pipelines have been parsed.
+						                sync_worker_memory_context(PARENT_PIPELINE_MEMORY_CONTEXT);
+
+						                dependencies.clear();
+						                std::swap(*parsed_parents, dependencies);
```
HansKristian-Work, line 2831: It seems flawed to allocate parsed_parents on the outside. We should be clearing parsed parents every iteration, since:

- parents holds the total accumulation of all parsed pipelines
- parsed_parents is an intermediate buffer which holds the pipelines we pushed this iteration

After an iteration is done, parents effectively consumed the information in parsed_parents, so we can safely clear it. It also avoids exploding parse time since we will be parsing the same pipeline for new parents over and over and over. By moving parsed_parents into this lambda and clearing it every iteration, I think we avoid that wart.

KonstantinSeurer: parsed_parents should already be cleared every iteration by this line:

```
dependencies.clear();
std::swap(*parsed_parents, dependencies);
```

Keeping parse time linear was the reason I introduced parsed_parents in the first place. I allocate parsed_parents outside the lambda, because the first iteration is done in a prior pass.

cli/fossilize_replay.cpp
```diff
@@ -2840,36 +2869,72 @@ struct ThreadedReplayer : StateCreatorInterface
...
+					                for (auto &parent : ordered_parents)
+					                {
+						                if (!parent.info)
+							                continue;
+
+						                if (!derived_work_item_is_satisfied(parent, pipelines))
+							                continue;
+
+						                uint32_t depth = parents_depth.at(parent.hash);
+						                if (prev_depth != depth) {
```
HansKristian-Work, line 2917: Brace

cli/fossilize_replay.cpp
HansKristian-Work: Right

HansKristian-Work (approved):
LGTM. Quite difficult to know 100% it's correct given the awkward concurrency going on here, but ...

PR #259, opened 2024-11-06
Add maintenance7 support for dynamic descriptors

djdeath:
The limits got more precise in maintenance7 and that is helping replay some traces that would otherwise not replay on Anv. 

Our limit is 16 dynamic descriptor in total, but we had to expose 8 for ubo/ssbo in the legacy limit.

### DIFF
--- a/cli/fossilize_feature_filter.cpp
+++ b/cli/fossilize_feature_filter.cpp
@@ -1896,10 +1896,22 @@ bool FeatureFilter::Impl::descriptor_set_layout_is_supported(const VkDescriptorS
 
 	if (pool_is_update_after_bind)
 	{
-		if (counts.ubo_dynamic > props.descriptor_indexing.maxDescriptorSetUpdateAfterBindUniformBuffersDynamic)
-			return false;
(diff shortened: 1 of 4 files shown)


cli/fossilize_feature_filter.cpp
```diff
@@ -1896,10 +1896,19 @@ bool FeatureFilter::Impl::descriptor_set_layout_is_supported(const VkDescriptorS
 
 	if (pool_is_update_after_bind)
 	{
-		if (counts.ubo_dynamic > props.descriptor_indexing.maxDescriptorSetUpdateAfterBindUniformBuffersDynamic)
-			return false;
-		if (counts.ssbo_dynamic > props.descriptor_indexing.maxDescriptorSetUpdateAfterBindStorageBuffersDynamic)
-			return false;
+		if (features.maintenance7.maintenance7) {
```
HansKristian-Work, line 1899: Braces on next line, otherwise, seems fine.

HansKristian-Work (changes requested):

cli/fossilize_feature_filter.cpp
djdeath: Done.

HansKristian-Work (approved):

Issue #266, opened 2025-04-10
mac build provokes warnings


lunarpapillo:
A **macOS** build works correctly, but provokes warnings that look like they could be improved.  All of the warnings are instances of one of these three:
```
% cmake -H. -Bbuild -DFOSSILIZE_VULKAN_LAYER=OFF -DFOSSILIZE_TESTS=OFF -DCMAKE_INSTALL_PREFIX=/Volumes/smb_share/ci-build-resources/mac/fossilize
...
% cmake --build build --target install
...
/Users/lunarg/.jenkins/workspace/update-fossilize/label/Mac/Fossilize/fossilize.cpp:114:2: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
        sprintf(str, "%016" PRIx64, value);
        ^
... 
/Users/lunarg/.jenkins/workspace/update-fossilize/label/Mac/Fossilize/fossilize.cpp:63:20: warning: unused function 'operator!=' [-Wunused-function]
static inline bool operator!=(const VkShaderModuleIdentifierEXT &a, const VkShaderModuleIdentifierEXT &b)
                   ^
... 
/Users/lunarg/.jenkins/workspace/update-fossilize/label/Mac/Fossilize/path.cpp:147:21: warning: unqualified call to 'std::move' [-Wunqualified-std-cast-call]
                        result.push_back(move(i));
                                         ^
                                         std::
```

PR #267, opened 2025-04-10
convert sprintf calls to snprintf

lunarpapillo:
Partly addresses:

    mac build provokes warnings
    https://github.com/ValveSoftware/Fossilize/issues/266

Many implementations have deprecated sprintf() in favor of snprintf(), including macOS with latest Xcode and Windows if _CRT_SECURE_NO_WARNINGS is set.

This change converts all uses of sprintf() to snprintf().  In most cases this is simply converting
    sprintf(x, ...)
to
    snprintf(x, sizeof(x), ...)

But there are a few more complicated constructions in fossilize_replay.cpp, fossilize_db.cpp, and layer/instance.cpp.

### DIFF
--- a/cli/fossilize_disasm.cpp
+++ b/cli/fossilize_disasm.cpp
@@ -876,7 +876,7 @@ static void print_help()
 static string uint64_string(uint64_t value)
 {
 	char str[17]; // 16 digits + null
-	sprintf(str, "%016" PRIx64, value);
+	snprintf(str, sizeof(str), "%016" PRIx64, value);
 	return string(str);
 }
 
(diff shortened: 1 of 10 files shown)


lunarpapillo:
> I am 99% this is purely a warning and just reporting to "fix it now before it causes issues down the road"

I'm 100% sure. :-)

Nothing is blocked (though we did have to alter the SDK build for this repository to turn off "warnings-are-errors" in order to unblock).

The warning that `sprintf()` is deprecated was stronger on some compilers than others, suggesting that support for the function will eventually be removed.  I'm not sure whether that's something to plan for, or a toothless threat.

I would not be concerned if smart people like y'all decided this wasn't ever going to be a real issue, and decided to close the PR without merge.

(I only made a PR because I'd gone through the effort to analyze the warnings after they first broke the SDK build, and I thought I understood at that point how to fix them, and thought it would be a shame to just drop them.)

PR #285, opened 2025-10-05
Add CLI fossilize-unmerge-db tool

JohnMega:
This PR adds a tool for unmerging databases after they have been merged into a single database. This is achieved by separating the databases with the `db_merge_separator` word when merging. This word is not considered by the `StreamArchieve` parser, so it is safe.


HansKristian-Work:
I still don't understand the point here. Why not just merge the DB in the first place if you're just going to unmerge them later? Which real-world problem does this solve? How would you deal with unmerge where two archives contain the same hash? Then it's not possible to reconstruct the originals anyway.

Without having a clear use case that I can see other users potentially finding meaningful (the other two PRs fit that criteria), I'm not inclined to merge support for it.

PR #286, opened 2025-10-08
Add overwrite_db_clear flag for DataBases

JohnMega:
Sometimes (for example, in the case of `fossilize-convert-db`), it can be convenient to constantly overwrite the results in the same folder. Currently, in the case of `DumbFolderDatabase`, the result of `fossilize-convert-db` is constantly overlaid with the result of the previous operation. This PR adds the ability to clear the previous result in `DumbFolderDatabase` when writing a new result using the `fossilize-convert-db` tool.

### DIFF
--- a/cli/fossilize_convert_db.cpp
+++ b/cli/fossilize_convert_db.cpp
@@ -55,7 +55,6 @@ int main(int argc, char *argv[])
 		print_help();
 		return EXIT_FAILURE;
 	}
-	
 	auto input_db = std::unique_ptr<DatabaseInterface>(create_database(argv[1], DatabaseMode::ReadOnly));
 	auto output_db = std::unique_ptr<DatabaseInterface>(create_database(argv[2], DatabaseMode::OverWrite));
 	if (!input_db || !input_db->prepare())
--- a/fossilize_db.cpp
+++ b/fossilize_db.cpp
@@ -558,9 +558,9 @@ struct DumbDirectoryDatabase : DatabaseInterface
 
 		dirent *next_file;
 
-		while ((next_file = readdir(dp)) != NULL)
+		while ((next_file = readdir(dp)) != nullptr)
 		{
-			if (strcmp(Path::ext(next_file->d_name).c_str(), "json"))
+			if (strcmp(Path::ext(next_file->d_name).c_str(), "json") != 0)
 				continue;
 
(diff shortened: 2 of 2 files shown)


cli/fossilize_convert_db.cpp
```diff
@@ -24,23 +24,38 @@
...
 {
-	LOGI("Usage: fossilize-convert-db input-db output-db\n");
+	LOGI("Usage: fossilize-convert-db input-db output-db\n"
+		"\t[--output-db-clear (only relevant for DumbDirectoryDatabase)]\n");
 }
 
 int main(int argc, char *argv[])
 {
+	bool overwrite_db_clear = false;
 	if (argc != 3)
```
HansKristian-Work: How will you even hit the code path to parse --output-db-clear now?

fossilize_db.cpp
```diff
@@ -701,6 +717,28 @@ struct DumbDirectoryDatabase : DatabaseInterface
...
 	unordered_set<Hash> seen_blobs[RESOURCE_COUNT];
+
+protected:
+	bool overwrite_db_clear()
+	{
+		DIR* dp = opendir(base_directory.c_str());
+		if (!dp)
+			return false;
+
+		dirent* next_file;
```
HansKristian-Work:
```suggestion
		dirent *next_file;
```

fossilize_db.cpp
```diff
@@ -701,6 +717,28 @@ struct DumbDirectoryDatabase : DatabaseInterface
 	string base_directory;
 	DatabaseMode mode;
 	unordered_set<Hash> seen_blobs[RESOURCE_COUNT];
+
+protected:
```
HansKristian-Work: Surely this can be private:?

fossilize_db.cpp
```diff
@@ -539,10 +550,36 @@ struct DumbDirectoryDatabase : DatabaseInterface
...
 
+	bool overwrite_db_clear()
+	{
+		DIR *dp = opendir(base_directory.c_str());
+		if (!dp)
+			return false;
+
+		dirent *next_file;
+
+		while ((next_file = readdir(dp)) != NULL)
```
HansKristian-Work:
```suggestion
		while ((next_file = readdir(dp)) != nullptr)
```

PR #304, opened 2026-06-12
Handle VkPipelineCreateFlags2 in disasm.

doitsujin:
Fixes a RADV crash when trying to dump ISA from DXVK pipelines.

Not sure if there's a nicer way to do this in the code base, this whole `const_cast` thing seems incredibly hacky.

### DIFF
--- a/cli/fossilize_disasm.cpp
+++ b/cli/fossilize_disasm.cpp
@@ -483,11 +483,11 @@ struct DisasmReplayer : StateCreatorInterface
 	}
 
 	template<typename T>
-	static void add_pipeline_flags(const T* create_info, VkPipelineCreateFlags2 flags)
+	static void add_pipeline_flags(const T *create_info, VkPipelineCreateFlags2 flags)
 	{
 		auto next = reinterpret_cast<const VkBaseInStructure *>(create_info->pNext);
 
-		while (next->sType != VK_STRUCTURE_TYPE_PIPELINE_CREATE_FLAGS_2_CREATE_INFO)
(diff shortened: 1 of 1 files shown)


cli/fossilize_disasm.cpp
```diff
@@ -482,6 +482,25 @@ struct DisasmReplayer : StateCreatorInterface
 		       !filter_raytracing.empty() || !filter_modules.empty();
 	}
 
+	template<typename T>
+	static void add_pipeline_flags(const T* create_info, VkPipelineCreateFlags2 flags)
+	{
+		auto next = reinterpret_cast<const VkBaseInStructure *>(create_info->pNext);
+
+		while (next->sType != VK_STRUCTURE_TYPE_PIPELINE_CREATE_FLAGS_2_CREATE_INFO)
```
HansKristian-Work:
```suggestion
		while (next && next->sType != VK_STRUCTURE_TYPE_PIPELINE_CREATE_FLAGS_2_CREATE_INFO)
```

cli/fossilize_disasm.cpp
```diff
@@ -482,6 +482,25 @@ struct DisasmReplayer : StateCreatorInterface
 		       !filter_raytracing.empty() || !filter_modules.empty();
 	}
 
+	template<typename T>
+	static void add_pipeline_flags(const T* create_info, VkPipelineCreateFlags2 flags)
```
HansKristian-Work:
```suggestion
	static void add_pipeline_flags(const T *create_info, VkPipelineCreateFlags2 flags)
```

Issue #303, opened 2026-06-10
[Enhancement] show status of the current shader compilation


miguelccoelho:
**As a** desktop user,
**I want to** verify the status of shader compilation
**So that I** can get some expectations of remaining work, time, and potentially decide to stop the processes.

One of the things that I believe makes people extra annoyed at fossilized taking a long time to compile shaders, is the fact that once it starts, there's no way to verify what's going on, expected runtime, or even a percentage on total work...

Basically, a simple improvement would be allow a user to even know what's going on in the first place. 
What shaders are being compiled? 
How many files? 
How long on average do they take?

Even if possible, generate a historical database of time taken for shader compilation for each game, so we can get an average expected time per job.

Help us understand, and that may even allow users to detect leaks/stuck jobs.

HansKristian-Work:
The --progress output has all that information, e.g.:

```
Fossilize INFO: =================
Fossilize INFO:  Progress report:
Fossilize INFO:    Overall 8462 / 112591
Fossilize INFO:    Parsed graphics 20749 / 61440, failed 28102, cached 0
Fossilize INFO:    Parsed compute 0 / 0, failed 0, cached 0
Fossilize INFO:    Decompress modules 13326 / 17226, skipped 0, failed validation 0, missing 0
Fossilize INFO:    Compile graphics 6847 / 61440, skipped 28102, cached 0
Fossilize INFO:    Compile compute 0 / 0, skipped 0, cached 0
Fossilize INFO:    Compile raytracing 0 / 0, skipped 0, cached 0
Fossilize INFO:    Clean crashes 0
Fossilize INFO:    Dirty crashes 0
Fossilize INFO: =================
```

but it's not really exposed in the Steam UI I assume? That's not really something Fossilize can solve on its own though.

Issue #302, opened 2026-05-25
Process fossilize_repla fail to be killed (zombie process) when trying to quit games


gregorywaynepower:
[steam_logs.txt](https://github.com/user-attachments/files/28228692/steam_logs.txt)

On Proton 11 (Beta), when trying to close Arc Raiders through the ingame menu, the game fails to fully close and I have seen have `fossilize_repla` processes that are unkillable. If there's any further information I need to gather, feel free to let me know.

I have used the following runtime options: `PROTON_LOG=1 SDL_VIDEO_DRIVER=x11 PROTON_ENABLE_WAYLAND=1 PROTON_USE_NTSYNC=1 gamemoderun %command%`. If some of these are unnecessary, feel free to let me know.

These zombie processes are visible upon launch:

```
[nix-shell:~]$ ps -e | grep defunct
   4236 ?        00:00:09 fossilize_repla <defunct>
   4237 ?        00:00:09 fossilize_repla <defunct>
   4238 ?        00:00:11 fossilize_repla <defunct>
   4239 ?        00:00:08 fossilize_repla <defunct>
   4240 ?        00:00:09 fossilize_repla <defunct>
   5643 ?        00:00:00 gpu_encoder_hel <defunct>
   5644 ?        00:00:00 gpu_encoder_hel <defunct>
   5690 ?        00:00:00 sd_voxin <defunct>
   5691 ?        00:00:00 sd_baratinoo <defunct>
   5693 ?        00:00:00 sd_cicero <defunct>
   5694 ?        00:00:00 sd_kali <defunct>
   5695 ?        00:00:00 sd_festival <defunct>
   5697 ?        00:00:00 sd_espeak-ng-mb <defunct>
   5808 ?        00:00:00 sd_openjtalk <defunct>
```

These are the zombie processes that are visible once I close the game and the game is running according to Steam:

```
[nix-shell:~]$ ps -eLF | grep defunct
gregory+    4236    2042    4236  0    2     0     0   2 23:34 ?        00:00:00 [fossilize_repla] <defunct>
gregory+    4237    2042    4237  0    2     0     0  12 23:34 ?        00:00:00 [fossilize_repla] <defunct>
gregory+    4238    2042    4238  0    2     0     0   6 23:34 ?        00:00:00 [fossilize_repla] <defunct>
gregory+    4239    2042    4239  0    2     0     0  10 23:34 ?        00:00:00 [fossilize_repla] <defunct>
[...]

Issue #288, opened 2025-12-01
104GB RAM Usage during "stuck" Processing Vulkan Shaders at 100%


hjpaul7:
#### Your system information

* Steam client version (build number or date): 1764106918
* Distribution (e.g. Ubuntu): Arch
* Opted into Steam client beta?: Yes
* Have you checked for system updates?: Yes
* Steam Logs: [steam-logs.tar.gz](https://github.com/user-attachments/files/23843401/steam-logs.tar.gz)

* GPU: Nvidia

#### Please describe your issue in as much detail as possible:

Steam is spawning multiple processes leading to 104GB RAM usage during "Processing Vulkan Shaders" while being stuck at 100%. I observed this for about 10 minutes and watched it reach 123GB out of 124GB multiple times and froze the system. Processing Vulkan shaders never completed, just sat there at 100% with this insanely high RAM usage.

#### Steps for reproducing this issue:

1. Launch Spider-Man 2
2. Let Vulkan Shaders process
3. Observe RAM usage and system freeze

<img width="1693" height="730" alt="Image" src="https://github.com/user-attachments/assets/8a8c4a5e-4d86-4822-8303-2dfbb0c4ee30" />

<img width="913" height="533" alt="Image" src="https://github.com/user-attachments/assets/9c821faf-aeb8-431b-bc85-77048227e781" />

louzt:
Quick note on #288 from the RFC angle since I happened to be looking at this area.

[PR #305](https://github.com/ValveSoftware/Fossilize/pull/305) addresses the **process orphaning** family: workers that survive Steam/Proton parent death and keep compiling in the background. The 104GB RAM case you're hitting looks like a different pattern — compilation state / VM growth where the replayer keeps allocating without releasing, even though the process tree is clean.

A few things that would help triage independently:

- `ps -eo pid,ppid,pcpu,pmem,vsz,rss,stat,cmd --sort=-rss | grep fossilize` while the behavior is happening — if `STAT` ends in `R` for one worker and that worker's `VSZ` is climbing into the 100GB range, that's the leaking-worker pattern.
- `nvidia-smi` simultaneously — if VRAM is pinned at high %, the GPU side is also leaking.
- `journalctl -k --since "5 min ago" | grep -i oom` — OOM events around the same window would confirm the working-set expansion.

[...]

Issue #279, opened 2025-08-22
Excessive CPU Usage when no game is running


alkaris2:
I'm noting an issue with `fossilize_replay` doesn't properly exit after a game has closed down, it'll peg the CPU usage to 100% on 2 cores on Linux, I'll have played a game for a couple hours maybe and quit the game, only to come back later to find CPU usage is still excessively high, because I can hear the fans ramped up more than they should be when not running a game or running any computationally intensive applications. I can see it's usage in `htop` having CPU utilization being at 100%, I can see it has 2 threads running the same thing with shadow processes, that even if you kill one of them, it doesn't kill the other and has to be killed separately. I feel this might need a kill script after a game has exited that looks for `fossilize_replay` running to kill it.

Possible related issues: #247 , #84 

```YAML
Computer Information:
Manufacturer: ASUSTeK COMPUTER INC.
Model: PRIME Z270-A
Form Factor: Desktop
No Touch Input Detected
Processor Information:
CPU Vendor: GenuineIntel
CPU Brand: Intel(R) Core(TM) i7-7700K CPU @ 4.20GHz
CPU Family: 0x6
CPU Model: 0x9e
CPU Stepping: 0x9
CPU Type: 0x0
Speed: 4500 MHz
8 logical processors
4 physical processors
Hyper-threading: Supported
FCMOV: Supported
SSE2: Supported
SSE3: Supported
SSSE3: Supported
SSE4a: Unsupported
SSE41: Supported
SSE42: Supported
AES: Supported
AVX: Supported
AVX2: Supported
AVX512F: Unsupported
AVX512PF: Unsupported
AVX512ER: Unsupported
AVX512CD: Unsupported
AVX512VNNI: Unsupported
SHA: Unsupported
CMPXCHG16B: Supported
LAHF/SAHF: Supported
PrefetchW: Unsupported
BMI1: Supported
BMI2: Supported
F16C: Supported
FMA: Supported
Operating System Version:
"Manjaro Linux" (64 bit)
Kernel Name: Linux
Kernel Version: 6.15.9-2-MANJARO
X Server Vendor: The X.Org Foundation
X Server Release: 12401008
X Window Manager: KWin
Steam Runtime Version: steam-runtime_1.0.20250519.130917
Client Information:
Version: 1751405894
Browser GPU Acceleration Status: Enabled
Browser Canvas: Enabled
Browser Canvas out-of-process rasterization: Enabled
[...]

PR #305, opened 2026-07-24
cli & replayer: Prevent Linux worker process orphaning with race-free PR_SET_PDEATHSIG

### LINKED ISSUE #247: Why is fossilize always running?
I have a 12 core Ryzen 5900x and only a few games on my computer, but it seems that steam is *constantly* running fossilize, the shader cache folder is a mere 2.4GB in size.

How on earth does it take my computer 28mins to compile a fully featured Linux kernel, but after rebooting, Steam spends *hours* processing who knows what?

I just cannot fathom this order of magnitude of required processing, Especially considering steam never even bothers to update my games, so I cannot fathom why It would need to spend so long processing shaders from World of Tanks and Apex Legends (and Noita/Factorio if that even counts as OpenGL titles).

louzt:
## Background

When a Steam/Proton parent exits abruptly (SIGKILL, crash, game exit), the
forked `fossilize_replay` / `ExternalReplayer` worker can end up reparented
to a subreaper (PID 1 in plain hosts, `pv-adverb` inside `pressure-vessel` /
`srt-bwrap`) and keep compiling Vulkan SPIR-V in the background.

Two upstream fixes already exist on this path:

- **#96 (2020, HansKristian-Work, merged)**: `close_fd` sentinel + `setpgid`
  + `killpg(SIGKILL)` in the replayer master. Handles the case where the
  spawned worker shares the parent's process group.
- **#233 (Windows-only)**: Job Objects with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`.

This PR complements #96 by binding the worker to a kernel-level death
signal (`prctl(PR_SET_PDEATHSIG, SIGKILL)`) so the cleanup also fires when
the worker ends up in a different process group (the CLI tool path used
when `inherit_process_group=true`, see `cli/fossilize_replay.cpp`). It also
adds explicit `--device-pci-vendor` / `--device-pci-device` selection so
hybrid laptops can pin shader compilation to the target dGPU instead of
falling through to the iGPU.

Closes: #247, #264, #279, #302.

#288 (104 GB RAM growth) and #303 (status flag) are tracked separately
and are not in scope of this PR — see "Out of scope" below.

---

## Relationship with PR #96

```mermaid
flowchart TD
    subgraph PR96["PR #96 (HansKristian-Work, 2020) - User-Space Group Cleanup"]
        P1["Worker spawned"] --> P2{"inherit_process_group?"}
        P2 -- "false (default)" --> P3["close_fd pipe + setpgid()"]
        P3 --> P4["epoll sentinel in master"]
        P4 --> P5{"Parent dies?"}
        P5 -- "Yes, gracefully" --> P6["killpg(SIGKILL) in user-space"]
        P5 -- "SIGKILL / OOM / crash" --> P7["Master code never runs, no killpg, worker orphaned to PID 1"]
        P2 -- "true (CLI path)" --> P8["Pipe and setpgid bypassed, no user-space cleanup at all"]
    end

    subgraph PR305["PR #305 (Louzt, 2026) - Kernel-Level Invariant"]
        Q1["Worker spawned"] --> Q2["prctl(PR_SET_PDEATHSIG, SIGKILL)"]
        Q2 --> Q3["getppid() race-free gate"]
        Q3 --> Q4["Kernel delivers SIGKILL on parent death, any pgroup or namespace"]
    end

    P7 -. "Orphaned to PID 1 or pv-adverb" .-> Q1
    P8 -. "Same orphan path" .-> Q1
```

PR #96 works while control returns to the replayer epoll loop. It cannot
cover three cases that PR #305 now does:

1. **Involuntary parent death (`SIGKILL`, Proton panic, OOM killer)**:
   the master code path is interrupted before `killpg` is reached, so the
   worker outlives the parent.
2. **`inherit_process_group=true` (CLI tool path, see
   `cli/fossilize_replay.cpp`)**: the `close_fd` pipe is only created when
   `!options.inherit_process_group`. When the CLI inherits the process
   group, no user-space cleanup runs at all.
3. **Container namespaces (`pressure-vessel` / `srt-bwrap`)**: workers
   reparented to `pv-adverb` (the container subreaper) escape the parent's
   process group, so neither `killpg` nor `setpgid` reaches them.

PR #305 binds each worker to a kernel death signal after `fork()` instead
of relying on user-space event loops, which closes all three gaps.

---

### Linux worker cleanup (`fossilize_replay_linux.hpp`, `fossilize_external_replayer_linux.hpp`)

After `fork()` in both the CLI replayer master and the `ExternalReplayer`
worker path:

```c
if (prctl(PR_SET_PDEATHSIG, SIGKILL) != 0)
{
    LOGE("Failed to set PR_SET_PDEATHSIG, aborting child startup.\n");
    _exit(EXIT_FAILURE);
}
if (getppid() != parent_pid)
    _exit(EXIT_FAILURE);
```

The `getppid()` re-check is the race-free gate: between `fork()` returning
and `prctl` taking effect, the parent could already have died and been
reaped. If `getppid() != parent_pid`, the parent is gone and we abort
cleanly instead of running unparented.

### Multi-GPU selection (`cli/device.cpp`, `cli/fossilize_replay.cpp`, `fossilize_external_replayer.cpp/hpp`)

New flags `--device-pci-vendor` and `--device-pci-device` (hex, e.g.
`0x10de` for NVIDIA). Selection is gated on `vendorID`; the optional
`deviceID` narrows it to a specific GPU. If the filter is set but no GPU
matches, `LOGW` warns and falls back to default selection instead of
silently picking the wrong device.

The predicate lives in `cli/device.hpp::VulkanDevice::pci_filter_matches`
so both the CLI selection loop and the unit test exercise the same logic.

---

## Verification

All 8 ctest pass (1.5 s on this host):

```
1/8 fossilize-system-test ............ Passed
2/8 varint-system-test ............... Passed
3/8 application-info-filter-test ..... Passed
4/8 object-cache-test ................ Passed
5/8 feature-filter-test .............. Passed
6/8 linux-futex-test ................. Passed
7/8 orphan-prevention-test ........... Passed
8/8 gpu-selection-test ............... Passed
```

- `orphan-prevention-test` forks an intermediate parent, kills it, and
  asserts the grandchild is reaped within 2 s (bounded poll, 200 × 10 ms).
- `gpu-selection-test` runs `pci_filter_matches` against a synthetic
  catalogue of NVIDIA / AMD / Intel IDs and checks vendor-only,
  vendor+device, mismatched device, and unset-filter cases.

The change is Linux-only; Windows behavior is unchanged and still
governed by Job Objects (#233).

---

## Out of scope

- Vulkan UMA / VRAM reclamation timing on SIGKILL — relies on the
  kernel's `drm_release()` path the same way every other Vulkan process
  does. Not benchmarked here.
- macOS — `prctl` is Linux-only. macOS still uses the PR #96 process-group
  cleanup.
- `#288` (104 GB RAM growth): separate pathological pattern — looks like
  compilation state / VM growth rather than orphaning. This PR does not
  address it; a separate report with `ps` / `nvidia-smi` output would help
  triage.
- `#303` (progress flag): enhancement request, not a bug fix. Not in scope
  of this PR.
- The "% CPU" numbers reported in the original PR description were not
  measured in this PR and have been removed.


HansKristian-Work:
This overall has a hint of LLM spam that I don't want to engage with, but the prctl idea makes sense.

I merged the prctl part through a squashed down commit on master. For device filtering, that could be a different PR since it's completely unrelated, but the added test coverage is somewhat overzealous and not needed unless I ask for it.

louzt:
Thanks for pulling the `prctl` kernel signal approach onto `master`, glad it helps other users! :)

To provide context on why I structured the PR this way, i really spent time debugging this background process leak on my local hybrid laptop setup (iGPU + dGPU) alongside the community reports/issues open i found useful later on that day, because this sintomps are kinda uncomfortable to deal with as user. 

From an outsider perspective trying to deliver this as complete as i can like a production ready fix for end users/or to ppl like you to help me solve it better, so bundling the kernel teardown, PCI device filters, and synthetic tests felt like the most thorough and respectful approach understanding what the Company behind means and that the collaborators had to read.
Quiet a time went into tracing this Linux kernel behavior on my hardware, trying other strats, the usual when you care, not a blind LLM prompt dump spam and token burn and so on, because that would fix this problem while ago, which is why the "don't want to engage" remark caught me off guard, like cmon, im free to PR if i feel something useful on the most useful way my exp can prepare on behalf of my name. But its okey. We're all busy and tired of slop doing all to being less sloppy, and you are the filter. 

Im are glad to help providing at least a good minimal triage/poc, so being open-minded about cutting or splitting scope is totally understandable, i was doubting about the proposal reach before even the first commit, so, keeping future contributions strictly atomic/short makes complete sense for upstream maintenance here and i will do that if found other time someday. Here was just matter of me solving my problem until i realize was a good PR op to help others on the road as OpenS and my person asks, i didnt knew u will reject my test because u dont ask for them or like those extra LOC i put around. If useful for others to trim ideas, ok, if not, ok with the trim. The squashed down commit, +10. Use what u really need on your edge. 

That said, get it. And sorry for being to -vvvv verbose on the PR body and here haha, just personality preferences (with and without LLM) i guess. Gonna continue outside. Very happy to see my laptop efficient on this edge case. Have a good day mate. Thanks for your time for review and merge.

HansKristian-Work:
Sorry, didn't mean to sound dismissive. When you're the filter for this stuff long enough, it wears you down and mistakes are made. Thanks for the prctl fix at least.

PR #308, opened 2026-07-30
Add a system to determine what a Fossilize bucket corresponds to

HansKristian-Work:
For future captures, a bucket.json is dumped alongside a bucket folder to make it possible to tie a bucket hash with a configuration.

For existing captures, a python script is added that can reverse a bucket.json file based on the filter.json and some a-priori knowledge about DXVK and vkd3d-proton.

### DIFF
--- a/fossilize.cpp
+++ b/fossilize.cpp
@@ -47,6 +47,7 @@
 #include "rapidjson/writer.h"
 using namespace rapidjson;
 
+using Allocator = RAPIDJSON_DEFAULT_ALLOCATOR;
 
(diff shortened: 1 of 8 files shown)


louzt:
nice additions! dumping `bucket.json` alongside the capture folder makes inspecting pipeline configs trivial, and `bucket-reverse.py` is a super clean approach to reverse those 64-bit hashes across multi-vendor setups.

solid QoL update for debugging captures. updating my edge builds with this, thanks for deep dive onto those topics Hans!

PR #314, opened 2026-08-12
fossilize-prune: drop unlinked library pipelines

mjanes-work:
fossilize-prune selects all pipelines created with VK_PIPELINE_CREATE_LIBRARY_BIT_KHR, as they may need to be pulled in from another pipeline.

DXVK uses VK_EXT_graphics_pipeline_library heavily.  As a result, fossilize-prune with --filter-graphics may select thousands of pipelines for a modern AAA title rendered via Proton.

Keep a list of the requested pipelines, and only output the entries that are required by them.


HansKristian-Work (approved):
Seems plausible.

